PR #1, opened 2019-03-11
Provide Node Client Library

k2la:
### Motivation
- Make it easier for Nodejs users to develop apps by using Pulsar.

Node client library implemented on C++ client library wrapped by [N-API](https://github.com/nodejs/node-addon-api).

This PR implements **producer/consumer** as a minimum of function by using N-API.

We'll add future tasks (Reader, Authentication, Test, CI/CD and so on) to ISSUE.

Authors: @k2la, @hrsakai, @massakam, @yumochiz, @nkurihar

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,25 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+
+BasedOnStyle: Google
+IndentWidth: 4
+ColumnLimit: 110
+SortIncludes: false
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterEnum: true
--- a/.clang-format
+++ b/.clang-format
@@ -17,7 +17,7 @@
 
 
 BasedOnStyle: Google
-IndentWidth: 4
+IndentWidth: 2
 ColumnLimit: 110
 SortIncludes: false
 BreakBeforeBraces: Custom
--- a/README.md
+++ b/README.md
@@ -23,6 +23,10 @@
 
 The Pulsar Node.js client can be used to create Pulsar producers and consumers in Node.js.
 
+## Compatibility
+
+This Node.js client is developed and tested using Apache Pulsar 2.3.0
+
 ## Requirements
 
 Pulsar Node.js client library is based on the C++ client library. Follow the instructions for
--- a/README.md
+++ b/README.md
@@ -1,1 +1,51 @@
-# Pulsar Node Client
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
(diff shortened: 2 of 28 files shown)


.clang-format
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+
+BasedOnStyle: Google
+IndentWidth: 4
```
sijie: it seems that we are using 4 spaces for clang and 2 spaces for node.

shall we make them consistent? or 2 spaces is preferred in node community?

README.md
```diff
@@ -1 +1,51 @@
...
+
+-->
+
+# Pulsar Node.js client library
+
+The Pulsar Node.js client can be used to create Pulsar producers and consumers in Node.js.
+
+## Requirements
+
+Pulsar Node.js client library is based on the C++ client library. Follow the instructions for
```
sijie: I would suggest add a section `Compatibility`. So people would have an expectation on what eversion of this node js client was developed.

E.g. `This NodeJS client is developed and tested using Apache Pulsar Client x.y.z`.

.clang-format
k2la: We'll make spaces consistent because some `*.cc` in [nodejs/node](https://github.com/nodejs/node) use 2 spaces.

README.md
merlimat: I still believe we should aim for a single release version. Otherwise things will get confusing on multiple fronts. Eg. from ASF perspective there's only one project "pulsar", hence I don't know how "partial" releases would work there.

examples/producer.js
```diff
@@ -0,0 +1,51 @@
...
+
+const Pulsar = require('../index.js');
+
+(async () => {
+  // Create a client
+  const clientConfig = {
+    serviceUrl: 'pulsar://localhost:6650',
+    operationTimeoutSeconds: 30,
+  };
+  const client = new Pulsar.Client(clientConfig);
```
merlimat: to make example more concise we could just pass the maps inline for all the configs: 

```js
    operationTimeoutSeconds: 30,
  });
```

k2la: Good. We fixed it.

examples/producer.js
```diff
@@ -0,0 +1,51 @@
...
+  // Send messages
+  const results = [];
+  for (let i = 0; i < 10; i += 1) {
+    const msg = `my-message-${i}`;
+    results.push(producer.send({
+      data: Buffer.from(msg),
+    }));
+    console.log(`Sent message: ${msg}`);
+  }
+  await Promise.all(results);
```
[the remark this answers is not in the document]
k2la: Thank you. We'll make ISSUE about it and add later.

README.md
sijie: I don't suggest us doing partial releases here. my suggestion here is to have a section for the current state of master in the repo. when people go to this repo and fork it, he should be aware of what version of Pulsar to use.

nkurihar: I also don't suggest the partial releases.
By the way, if we don't do the partial release, should we set the version number of node.js client to same as apache/pulsar? (like 2.3.0)

sijie: > should we set the version number of node.js client to same as apache/pulsar? (like 2.3.0)

yes. I think we should do that.

k2la: We added `Compatibility` section and set the version number of node.js client to same as `apache/pulsar`

examples/producer.js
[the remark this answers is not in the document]
nkurihar: I just created the issue:
https://github.com/apache/pulsar-client-node/issues/2
On another PR we will try to implement.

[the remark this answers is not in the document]
merlimat: 👍

README.md
merlimat: Maybe we can use the same Maven convention for versioning? eg: 2.4.0-SNAPSHOT

k2la: We can use like eg. set 2.4.0-SNAPSHOT.

PR #27, opened 2019-04-22
Add scripts for tests in docker container

hrsakai:
`docker-tesh.sh` script executes tests in following steps.
1. Run docker container
2. Download/install cpp client pkg
3. Download pulsar distribution pkg and run standalone 
4.  npm install  and run tests
5. Stop standalone


pulsar-test-service-start.sh
```diff
@@ -0,0 +1,45 @@
...
+# specific language governing permissions and limitations
+# under the License.
+#
+
+set -e
+
+ROOT_DIR=$(git rev-parse --show-toplevel)
+cd $ROOT_DIR
+
+VERSION="${VERSION:-2.3.1}"
```
merlimat: We should define the version in one file and refer to it from multiple places

hrsakai: I added `pulsar-version.txt` file and defined the version in that file.

PR #31, opened 2019-06-11
Implement Reader

k2la:
- Add `createReader` to Client Class
- Add Reader Class
  - readNext
  - hasNext
  - close
- Add ReaderConfig Class


src/Reader.cc
```diff
@@ -0,0 +1,185 @@
...
+    ReaderReadNextWorker *wk = new ReaderReadNextWorker(deferred, this->cReader, timeout.Int64Value());
+    wk->Queue();
+  }
+  return deferred.Promise();
+}
+
+Napi::Value Reader::HasNext(const Napi::CallbackInfo &info) {
+  int value = 0;
+  pulsar_result result = pulsar_reader_has_message_available(this->cReader, &value);
+  if (result != pulsar_result_Ok || value != 1) {
```
massakam: Shouldn't we throw an exception if the result is not ok?

k2la: Yes. I think we should throw it.
So, I fixed same as https://github.com/apache/pulsar/blob/master/pulsar-client-go/pulsar/c_reader.go#L163-L169

massakam:
A binary file is committed.
```
.vscode/ipch/cc6423c2fd8ec533/mmap_address.bin
```

k2la:
I deleted `.vscode` dir

PR #35, opened 2019-06-13
Move Libraries in Perf to devDependencies

k2la:
- Moved  following libraries in perf to devDependencies:
  - commander
  - delay
  - hdr-histogram-js
  - lodash
- Added `.eslintignore`
  - Ignore `perf/*` because they are caught in a problem that they don't exist in `dependencies` by eslint


hrsakai:
Could you update `package-lock.json`?

k2la:
@hrsakai updated it.

Issue #39, opened 2019-07-26
double free or corruption (out): 0x00007f85abffad00


freeart:
Pulsar 2.4 runs in docker on the remote machine 
The project has three readers like this. 
Common client between them and 3 local readers.
Each of them reads own topic.
```js
//common client
scope.client = new Pulsar.Client({
		serviceUrl: 'pulsar://xx.xx.xx.xx:6650',
		operationTimeoutSeconds: 30
});
//different files
const reader = await scope.client.createReader({
	topic: 'persistent://public/default/xxx',
	startMessageId: Pulsar.MessageId.earliest()
});
async.forever(async () => {
	let msg = await reader.readNext();
	const msgId = msg.getMessageId().serialize()
	const payload = JSON.parse(msg.getData().toString());
	// async work...
	return Promise.resolve()
})
```
Get this error 3 times when new messages come to topic. Don't know what exactly was the cause
```
*** Error in `node /usr/src/apps/ms-log/index.js': double free or corruption (out): 0x00007f85abffad00 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x70bfb)[0x7f85bc7e2bfb]
/lib/x86_64-linux-gnu/libc.so.6(+0x76fc6)[0x7f85bc7e8fc6]
/lib/x86_64-linux-gnu/libc.so.6(+0x7780e)[0x7f85bc7e980e]
/usr/src/apps/ms-log/node_modules/pulsar-client/build/Release/Pulsar.node(_ZN9MessageIdD1Ev+0x15)[0x7f85b851cbe5]
/usr/src/apps/ms-log/node_modules/pulsar-client/build/Release/Pulsar.node(_ZN4Napi10ObjectWrapI9MessageIdE16FinalizeCallbackEP10napi_env__PvS5_+0x19)[0x7f85b8520e09]
node /usr/src/apps/ms-log/index.js[0x8d79a6]
node /usr/src/apps/ms-log/index.js(_ZN2v88internal13GlobalHandles32InvokeSecondPassPhantomCallbacksEPNS0_4ListINS1_22PendingPhantomCallbackENS0_25FreeStoreAllocationPolicyEEEPNS0_7IsolateE+0xa8)[0xddd9e8]
node /usr/src/apps/ms-log/index.js(_ZN2v88internal13GlobalHandles37PendingPhantomCallbacksSecondPassTask11RunInternalEv+0x179)[0xdddb79]
[...]

PR #41, opened 2019-08-05
[Issue #39]Add skip flag for cMessageId free

hrsakai:
It seems that [the issue](https://github.com/apache/pulsar-client-node/issues/39) occurs when gc releases `earliest(latest) messageId`.
These messageIds shouldn't be released.

In go client library, it doesn't call `pulsar_message_id_free()` if messageid is earliest or latest.
https://github.com/apache/pulsar/blob/d5036eaefcba42e6c9d159ffb14bee597d13380f/pulsar-client-go/pulsar/c_message.go#L198-L206
https://github.com/apache/pulsar/blob/d5036eaefcba42e6c9d159ffb14bee597d13380f/pulsar-client-go/pulsar/c_message.go#L169-L177

### DIFF
--- a/src/MessageId.h
+++ b/src/MessageId.h
@@ -41,6 +41,7 @@ class MessageId : public Napi::ObjectWrap<MessageId> {
  private:
   static Napi::FunctionReference constructor;
   pulsar_message_id_t *cMessageId;
+  bool skipcMessageIdFree = false;
 
   Napi::Value ToString(const Napi::CallbackInfo &info);
 };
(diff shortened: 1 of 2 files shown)


src/MessageId.h
```diff
@@ -41,6 +41,7 @@ class MessageId : public Napi::ObjectWrap<MessageId> {
  private:
   static Napi::FunctionReference constructor;
   pulsar_message_id_t *cMessageId;
+  bool skipcMessageIdFree = false;
```
massakam: I think skip**C**MessageIdFree is correct as a camel case variable.

hrsakai: Fixed it.

PR #47, opened 2019-08-21
added note for pulsar-client-dev rather than pulsar-client

daghdha:
just changing the README so that others don't run into the same problem I did.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,8 @@ Pulsar Node.js client library is based on the C++ client library. Follow the ins
 [Deb](https://pulsar.apache.org/docs/en/client-libraries-cpp/#deb) or
 [Homebrew packages](https://pulsar.apache.org/docs/en/client-libraries-cpp/#macos).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,6 +35,8 @@ Pulsar Node.js client library is based on the C++ client library. Follow the ins
 [Deb](https://pulsar.apache.org/docs/en/client-libraries-cpp/#deb) or
 [Homebrew packages](https://pulsar.apache.org/docs/en/client-libraries-cpp/#macos).
 
+(Note: you will need to install the pulsar-client-dev library, not the pulsar-client library.)
```
massakam: Since pulsar-client-dev depends on pulsar-client, users will need to install both packages. So, I think it is better to add the following note:

> Note: you will need to install not only the pulsar-client library but also the pulsar-client-dev library

daghdha: ok.  I'll change that and make another PR

PR #81, opened 2020-03-11
Support seting topic schema when creating producers/consumers

lcartercondon:
Added class to allow setting topic schema when creating
producers/consumers


src/ProducerConfig.cc
```diff
@@ -33,6 +33,11 @@ static const std::string CFG_COMPRESS_TYPE = "compressionType";
 static const std::string CFG_BATCH_ENABLED = "batchingEnabled";
 static const std::string CFG_BATCH_MAX_DELAY = "batchingMaxPublishDelayMs";
 static const std::string CFG_BATCH_MAX_MSG = "batchingMaxMessages";
+static const std::string CFG_SCHEMA = "schema";
+static const std::string CFG_SCHEMA_TYPE = "schemaType";
+static const std::string CFG_SCHEMA_NAME = "name";
+static const std::string CFG_SCHEMA_SCHEMA = "schema";
+static const std::string CFG_SCHEMA_PROPS = "properties";
```
mkaufmaner:
```suggestion
static const std::string CFG_SCHEMA = "schema";
```

src/SchemaInfo.h
```diff
@@ -0,0 +1,45 @@
...
+#include <pulsar/c/consumer_configuration.h>
+
+class SchemaInfo {
+ public:
+  SchemaInfo(const Napi::Object &schemaInfo);
+  ~SchemaInfo();
+  // pulsar_schema_type GetSchemaType();
+  // std::string GetName();
+  // std::string GetSchema();
+  // pulsar_string_map_t GetProperties();
```
mkaufmaner:
```suggestion
```

src/ProducerConfig.cc
lcartercondon: Yup, thanks for the heads up on cleanup.

lcartercondon: Fixed in 7a29a67

src/SchemaInfo.h
lcartercondon: Fixed in 7a29a67

massakam:
Even if we can set the schema information, the current Node.js client can handle message payloads only with Buffer type.
https://github.com/apache/pulsar-client-node/blob/cc79c67c3c330be0667203bf369b5de625eb0044/src/Message.cc#L136-L143
https://github.com/apache/pulsar-client-node/blob/cc79c67c3c330be0667203bf369b5de625eb0044/src/Message.cc#L85-L93

lcartercondon:
This simply allows the NodeJS client to specify the schema when creating the producer or consumer - it doesn't go to the same extent that, say, the Python client does where passing an object (to the producer's send) causes that object to get encoded using the schema info or where the subscriber decodes the message payload to a value using the schema info.

I'm planning on adding the automatic encode/decode parts in a similar fashion to how the Python client does it i.e. by wrapping the Producer/Consumer/Message bindings in language-native classes that add information regarding schema and support for automatic encode/decode, but I'm not quite sure yet of the exact implementation.

lcartercondon:
@massakam I think this is fine - any user of a schema'd topic is (with this implementation) still responsible for serializing to Buffer when sending messages or deserializing from Buffer into whatever schema when consuming messages.

massakam:
> I think this is fine - any user of a schema'd topic is (with this implementation) still responsible for serializing to Buffer when sending messages or deserializing from Buffer into whatever schema when consuming messages.

Once users implement such an application, isn't it a breaking change for them to support automatic encoding/decoding?
```js
const buffer = /* object serialization> */;
producer.send({
  data: buffer,
});
```
↓
```js
const obj = {
  key1: "val1",
  key2: "val2",
};
producer.send({
  data: obj,
});
```
If so, I think parts for automatic encoding / decoding should be added in this PR or as soon as possible.

lcartercondon:
> isn't it a breaking change for them to support automatic encoding/decoding?

This is a great point - I hadn't thought about it yet. I could work around it in `producer.send` by skipping the automatic encode if data is of a buffer type, or make automatic encode opt-in either during producer creation or on each send call.

It's definitely a bit more tricky for the message handling function in the consumer. Would it be reasonable to do something along the lines of what the Python lib does? It assigns the decoded `message.data` to a `message.value` attribute. Alternatively, could make the automatic decode opt-in during consumer creation.

massakam:
In the case of Python and Go libraries, it seems that binary data and deserialized data can be obtained from different fields of the message object. If the Node.js library follows this, it won't be a breaking change.
https://github.com/apache/pulsar/blob/d5036eaefcba42e6c9d159ffb14bee597d13380f/pulsar-client-go/pulsar/message.go#L47-L75

ludusrusso:
Any news about this? It would be great to have schemas in nodejs! Can I help to complete the PR?

lcartercondon:
> Any news about this? It would be great to have schemas in nodejs! Can I help to complete the PR?

I've not taken a look at this in quite some time unfortunately - I was working on this as part of related work at a prior employer. I'll take a look at this again this weekend, but cannot make any guarantees as to completing the requested changes.

Contributions that complete the requests are absolutely welcome, I believe all that's left is:
1. provide example documentation of how a NodeJS client would use this feature.

There has been some chatter in this PR re adding automatic encode/decode support, similar to how Python and Go client libs do it. Contributions that add that would be above and beyond but are absolutely welcome as well.

I'll update this weekend if I'm able to accomplish the requested changes and/or automatic encode/decode support feature request.

ludusrusso:
@lcartercondon hi, thanks for the reply.
If it's ok for you, I can continuo working on the two request: 
documentation and auto encode/decode, that wold be super helpful for my use case!

Alternatively, you can close this PR with docs and I can submit an new PR with audo encode/decode!

tschmidt64-SC:
Any updates on this? Seems like a really useful feature. As I understand it, without schema support, Pulsar SQL is basically non-functional without this feature. Am I correct in that understanding?

thomasechen:
@sijie Anu update on this request and it is really important to the world because numbers of nodejs developmer become more amd more

k2la:
@lcartercondon Could you please rebase to the master branch?

k2la (approved):

massakam:
I've resolved the conflict, but the current code fails to compile.

src/SchemaInfo.cc
```diff
@@ -0,0 +1,78 @@
...
+    for (int i = 0; i < size; i++) {
+      Napi::String key = arr.Get(i).ToString();
+      Napi::String value = propObj.Get(key).ToString();
+      pulsar_string_map_put(this->cProperties, key.Utf8Value().c_str(), value.Utf8Value().c_str());
+    }
+  }
+}
+
+void SchemaInfo::SetProducerSchema(pulsar_producer_configuration_t *cProducerConfiguration) {
+  pulsar_producer_configuration_set_schema_info(cProducerConfiguration, this->cSchemaType, this->name.c_str(),
```
massakam, line 69:
```suggestion
void SchemaInfo::SetProducerSchema(std::shared_ptr<pulsar_producer_configuration_t> cProducerConfiguration) {
  pulsar_producer_configuration_set_schema_info(cProducerConfiguration.get(), this->cSchemaType, this->name.c_str(),
```

src/SchemaInfo.cc
```diff
@@ -0,0 +1,78 @@
...
+                                                this->schema.c_str(), this->cProperties);
+}
+
+void SchemaInfo::SetConsumerSchema(pulsar_consumer_configuration_t *cConsumerConfiguration) {
+  pulsar_consumer_configuration_set_schema_info(cConsumerConfiguration, this->cSchemaType, this->name.c_str(),
```
massakam, line 74:
```suggestion
void SchemaInfo::SetConsumerSchema(std::shared_ptr<pulsar_consumer_configuration_t> cConsumerConfiguration) {
  pulsar_consumer_configuration_set_schema_info(cConsumerConfiguration.get(), this->cSchemaType, this->name.c_str(),
```

src/SchemaInfo.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include <napi.h>
+#include <pulsar/c/producer_configuration.h>
+#include <pulsar/c/consumer_configuration.h>
+
+class SchemaInfo {
+ public:
+  SchemaInfo(const Napi::Object &schemaInfo);
+  ~SchemaInfo();
+  void SetProducerSchema(pulsar_producer_configuration_t *cProducerConfiguration);
```
massakam, line 31:
```suggestion
  void SetProducerSchema(std::shared_ptr<pulsar_producer_configuration_t> cProducerConfiguration);
```

src/SchemaInfo.h
```diff
@@ -0,0 +1,41 @@
...
+  void SetConsumerSchema(pulsar_consumer_configuration_t *cConsumerConfiguration);
```
massakam, line 32:
```suggestion
  void SetConsumerSchema(std::shared_ptr<pulsar_consumer_configuration_t> cConsumerConfiguration);
```

massakam:
@lcartercondon Could you fix the compilation error?

nkurihar:
This change is merged as a new PR with fixing some compliation errors: 
https://github.com/apache/pulsar-client-node/pull/223

@lcartercondon Thank you for your contribution!

Issue #78, opened 2020-03-01
Error: Failed to send message: AlreadyClosed 


alijmlzd:
Hi there,
Why this example with 1 second timeout works fine:
```js
const client = new Pulsar.Client({
    serviceUrl: 'pulsar://localhost:6650',
})

const producer = await client.createProducer({
    topic: 'persistent://public/default/my-topic',
})

setTimeout(() => {
    producer.send({
        data: Buffer.from('My Message'),
    })
}, 1000)
```
But this example with a few more seconds timeout occurs an error?
```js
const client = new Pulsar.Client({
    serviceUrl: 'pulsar://localhost:6650',
})

const producer = await client.createProducer({
    topic: 'persistent://public/default/my-topic',
})

setTimeout(() => {
    producer.send({
        data: Buffer.from('My Message'),
    })
}, 9000)
```
Full error message:
```
You have triggered an unhandledRejection, you may have forgotten to catch a Promise rejection:
Error: Failed to send message: AlreadyClosed
```

alijmlzd:
Is there any kind of activity is there to solve this issue?
To be more specific, I run pulsar inside Docker like this [as described here](https://pulsar.apache.org/docs/v2.0.1-incubating/getting-started/docker/):
```bash
$ docker run -it \
  -p 6650:6650 \
  -p 8080:8080 \
  -v $PWD/data:/pulsar/data \
  apachepulsar/pulsar:2.0.0-rc1-incubating \
  bin/pulsar standalone
```
And here is the pulsar server logs:
```bash
18:33:03.492 [pulsar-io-50-4] INFO  org.apache.pulsar.broker.service.ServerCnx - New connection from /172.17.0.1:55506
18:33:03.607 [pulsar-io-50-4] INFO  org.apache.pulsar.broker.service.ServerCnx - [/172.17.0.1:55506][persistent://public/default/my-topic] Creating producer. producerId=0
18:33:03.722 [Thread-5] INFO  org.apache.pulsar.broker.service.ServerCnx - [/172.17.0.1:55506] persistent://public/default/my-topic configured with schema false
18:33:03.728 [Thread-5] INFO  org.apache.pulsar.broker.service.ServerCnx - [/172.17.0.1:55506] Created new producer: Producer{topic=PersistentTopic{topic=persistent://public/default/my-topic}, client=/172.17.0.1:55506, producerName=standalone-8-2, producerId=0}
[...]

PR #85, opened 2020-03-25
Create reference for client

yosiat:
Since we don’t have reference for the client, v8 can garbage collect this instance, which can cause the next code:

```

(async () => {
  // Create a client
    serviceUrl: 'pulsar://localhost:6650'
  });

  // Create a producer
    sendTimeoutMs: 30000,
    batchingEnabled: true,
  });

  let i =0;
  async function produceMessage() {
    i = i + 1;

    const msg = `my-message-${i}`;
    const ret = await producer.send({
      data: Buffer.from(msg),
    });
    console.log(`Sent message: ${msg}`);
    console.log(ret);

    setTimeout(produceMessage, 1000);
  }

  produceMessage();
})();
```

To close the connection suddenly and fail the producer.

Fixes: https://github.com/apache/pulsar-client-node/issues/78

### DIFF
--- a/src/Client.cc
+++ b/src/Client.cc
@@ -151,6 +151,7 @@ Client::Client(const Napi::CallbackInfo &info) : Napi::ObjectWrap<Client>(info) 
 
   this->cClient = pulsar_client_create(serviceUrl.Utf8Value().c_str(), cClientConfig);
   pulsar_client_configuration_free(cClientConfig);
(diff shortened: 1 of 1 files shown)


yosiat:
@hrsakai @sijie can you please review this PR ?

sijie:
@yosiat this is cool! Thank you for your contribution!

PR #106, opened 2020-07-20
Fix Cleanup Listener

k2la:
### Motivation
When the following code is executed, the node process is not finished in spite of consuming all messages and closing the consumer and client.

```

(async () => {
  });
  let received = 0;
  let finish;
  const finishPromise = new Promise((resolve) => {
    finish = resolve;
  });

    subscription: 'sub1',
    listener: (msg, msgConsumer) => {
      const data = msg.getData().toString();
      console.log(data);
      msgConsumer.acknowledge(msg);
      received += 1;
      if (received === 10) finish();
    },
  });

  await finishPromise;
  await consumer.close();
  await client.close();
})();
```

It seems that this happens because the registered listener function is not released.
https://github.com/apache/pulsar-client-node/blob/master/src/Consumer.cc#L307 
`this->listener->callback.Release();` is not executed because `this->listener` becomes `nullptr` when consumers are closed.

### Modification
Fix so that the registered listener function is released when the consumer is closed.

### DIFF
--- a/src/Consumer.cc
+++ b/src/Consumer.cc
@@ -289,9 +289,15 @@ class ConsumerCloseWorker : public Napi::AsyncWorker {
 
 void Consumer::Cleanup() {
   if (this->listener) {
-    this->Unref();
-    this->listener = nullptr;
+    this->CleanupListener();
   }
+}
+
+void Consumer::CleanupListener() {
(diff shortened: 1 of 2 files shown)


nkurihar:
test this please

PR #109, opened 2020-08-06
Bump the master version to 1.3.0

massakam:
I have already created `branch-1.2` to perform the release, so I will bump the master version to `1.3.0-rc.0`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "pulsar-client",
-  "version": "1.2.0-rc.0",
+  "version": "1.3.0-rc.0",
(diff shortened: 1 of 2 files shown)


lxcid:
v1.3.0-rc.0 seems to have been stuck for a while, can we release it. The 2 incomplete milestone seems to be new feature and not anything that are blockers.

lxcid:
The reason is that 1.2.0 have compilation issue and v1.3.0 have that fix I think. As v1.2.0 are using old `node-pre-gyp`

nkurihar:
@lxcid Sorry for delay of the release.
We're planning the next release in March.

PR #117, opened 2020-08-28
Add github actions

hrsakai:
Use github actions for precommit job.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -1,0 +1,13 @@
+name: Node.js
+on: [pull_request]
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code into the Node.js module directory
+        uses: actions/checkout@v2
+
+      - name: Test
+        run: |
+          ./docker-tests.sh


.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,13 @@
+name: Node.js
+on: [pull_request]
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
```
massakam, line 6: Is it not possible to use the `apachepulsar/pulsar-build` image directly?

okuryu: You can probably do that with self-hosted runners.
https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners

hrsakai: @massakam 
By default, it seems that only the virtual environments listed in the table on the following page can be used.
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

@okuryu 
Thank you for your comment.
I will confirm self-hosted runners.

hrsakai: I read the documents for self-hosted runners.
There are things to consider(security,server,etc), so we would like to consider self-hosted runners again in the future as needed.

nkurihar (approved):

massakam (approved):

Issue #11, opened 2019-03-14
Support CompressionZSTD


k2la:
Support ZSTD for compression.

Currently, about compression, `None`, `Zlib` and `LZ4`. are supported.
https://github.com/apache/pulsar-client-node/blob/master/src/ProducerConfig.cc#L49-L50

It is not in `client-cpp/include/pulsar/c/producer_configuration.h`. https://github.com/apache/pulsar/blob/79b88cf66eb2a4d337e21ade7f699c7d641b5c9a/pulsar-client-cpp/include/pulsar/c/producer_configuration.h#L40-L44

hrsakai:
Closed by https://github.com/apache/pulsar-client-node/pull/121

PR #121, opened 2020-09-12
Includes Support for ZSTD and SNAPPY Compression

savearray2:
A version bump to the latest version of the Pulsar C++ library (2.6.1) was required to make use of the compression types. As the latest version of the C++ library requires ```GLIBCXX_3.4.22```, the testing suite code was altered to automatically update the Docker image with the latest version of ```libstdc++6``` and ```gcc-4.9``` before executing the library tests.

Since this makes use of a later version of the C++ library, compatibility requirements will change, and should probably be reflected in ```README.md```.

Accommodates the issue created here: https://github.com/apache/pulsar-client-node/issues/11#issue-420827873


pulsar-version.txt
```diff
@@ -1 +1 @@
-2.5.0
+2.6.1
```
massakam, line 1: ZSTD and SNAPPY seem to be available from Pulsar v2.6.0.
https://github.com/apache/pulsar/pull/7014

tests/end_to_end.test.js
```diff
@@ -376,5 +376,53 @@ const Pulsar = require('../index.js');
...
+        operationTimeoutSeconds: 30,
+      });
+      expect(client).not.toBeNull();
+
+      const topic = 'persistent://public/default/produce-read-compression';
+      const producer = await client.createProducer({
+        topic,
+        sendTimeoutMs: 30000,
+        batchingEnabled: true,
+        compressionType: 'ZSTD',
```
massakam, line 392: Could you add a test for SNAPPY as well?

massakam:
Please resolve the conflict.

tests/end_to_end.test.js
massakam: Ah, it seems that sending a message fails when the compression type is set to SNAPPY. I think this is because the Docker image for building the DEB package doesn't include the SNAPPY library.
https://github.com/apache/pulsar/blob/master/pulsar-client-cpp/pkg/rpm/Dockerfile

So we shouldn't add a test for SNAPPY yet.

savearray2: I can see if I can update the docker container and include the tests.

massakam: I don't think SNAPPY is available unless the SNAPPY library is installed when the C++ client DEB package is built.
https://github.com/apache/pulsar/pull/4259/files#diff-bda8db832e446d092d38464123ed17bc
https://github.com/apache/pulsar/pull/4259/files#diff-7e441fcfcb20e339c71f7f8f48569579

So, even if you modify the Docker image, we cannot use SNAPPY immediately. We have to wait for the next release of the C++ package.

massakam: Created a pull-request to the apache/pulsar repository.
https://github.com/apache/pulsar/pull/8086

pulsar-version.txt
savearray2: Would you like me to change this to ```2.6.0``` instead of ```2.6.1```?

massakam: Yes. I think the C++ client used for testing should be the lowest version that this library works with.

massakam (approved):

PR #126, opened 2020-09-27
Adds getProducerName and getTopic to Producer

alanhoff:
It simply hooks up `pulsar_producer_get_topic` and `pulsar_producer_get_producer_name` so the JS client can have access to these values.

### DIFF
--- a/tests/producer.test.js
+++ b/tests/producer.test.js
@@ -57,12 +57,22 @@ const Pulsar = require('../index.js');
         })).rejects.toThrow('Failed to create producer: ConnectError');
       });
 
-      test('Producer Name', async () => {
+      test('Automatic Producer Name', async () => {
         const producer = await client.createProducer({
           topic: 'persistent://public/default/topic',
         });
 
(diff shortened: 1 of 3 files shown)


tests/producer.test.js
```diff
@@ -56,6 +56,24 @@ const Pulsar = require('../index.js');
           batchingEnabled: true,
         })).rejects.toThrow('Failed to create producer: ConnectError');
       });
+
+      test('Producer Name', async () => {
+        const producer = await client.createProducer({
+          topic: 'persistent://public/default/topic',
+        });
+
+        expect(typeof producer.getProducerName()).toBe('string');
```
hrsakai, line 65: @alanhoff 

Thank you for your contribution.

[nits]
Could you set producerName and check that string?

e.g.
```
          topic: 'persistent://public/default/topic',
          producerName: 'aaaaa'
        });

expect(producer.getProducerName()).toBe('aaaaa');
```

alanhoff: Fixed

hrsakai (approved):
LGTM

PR #127, opened 2020-09-27
feature: allow messages and producers to handle int64 sequences

### LINKED ISSUE #124: Allow BigInt/Buffer to be used as initialSequenceId and/or sequenceId
Hello there,

Right now the producer's sequence ID is limited to 4 bytes here:

- https://github.com/apache/pulsar-client-node/blob/master/src/ProducerConfig.cc#L76
- https://github.com/apache/pulsar-client-node/blob/master/src/Message.cc#L174

However we could allow 8 bytes integers if we allow producer/message to receive a [BigInt](https://nodejs.org/api/n-api.html#n_api_napi_get_value_bigint_int64) and/or a [8 bytes buffer](https://nodejs.org/api/buffer.html#buffer_buf_writebigint64le_value_offset).

I'm trying to match a dataset that uses 8 bytes integers for keys as my sequence ID, that's why I found this limitation :-)

alanhoff:
* Allows the initial sequence ID to be passed as a string: `client.createProducer({initialSequenceId: '1234'})` so JS can send 64 bits integers encoded as string.
* Allows the message sequence ID to be passed as a string `producer.send({sequenceId: '1234'})` so JS can send 64 bits integers encoded as string.
* Implements `Producer.getLastSequenceId` which returns `Napi::String` since `Napi::Number` cannot be used for 64 bits integers.
* Closes #124

### DIFF
--- a/src/Message.cc
+++ b/src/Message.cc
@@ -20,6 +20,7 @@
 #include "Message.h"
 #include "MessageId.h"
 #include <pulsar/c/message.h>
+#include <string>
 
 static const std::string CFG_DATA = "data";
 static const std::string CFG_PROPS = "properties";
@@ -174,6 +175,11 @@ pulsar_message_t *Message::BuildMessage(Napi::Object conf) {
   if (conf.Has(CFG_SEQUENCE_ID) && conf.Get(CFG_SEQUENCE_ID).IsNumber()) {
     Napi::Number sequenceId = conf.Get(CFG_SEQUENCE_ID).ToNumber();
     pulsar_message_set_sequence_id(cMessage, sequenceId.Int64Value());
+  }
+
+  if (conf.Has(CFG_SEQUENCE_ID) && conf.Get(CFG_SEQUENCE_ID).IsString()) {
+    std::string sequenceId = conf.Get(CFG_SEQUENCE_ID).ToString().Utf8Value();
+    pulsar_message_set_sequence_id(cMessage, std::strtoll(sequenceId.c_str(), 0, 10));
   }
 
   if (conf.Has(CFG_PARTITION_KEY) && conf.Get(CFG_PARTITION_KEY).IsString()) {
--- a/src/Producer.cc
+++ b/src/Producer.cc
@@ -22,6 +22,8 @@
 #include "Message.h"
 #include <pulsar/c/result.h>
 #include <memory>
+#include <string>
+
 Napi::FunctionReference Producer::constructor;
 
 void Producer::Init(Napi::Env env, Napi::Object exports) {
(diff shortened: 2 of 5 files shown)


alanhoff:
@hrsakai I've rebased my PR and it's ready for review 👍

hrsakai:
@alanhoff 
I can't reproduce the issue.
I set 5294967295(greater than 32 bits integers) to initialSequenceId, but sequence id looks correct.
Please tell me how to reproduce.
```
initialSequenceId = 5294967295
lastSequenceId = 5294967395 
```

code
```
const Pulsar = require('../index.js');

(async () => {
  // Create a client
    operationTimeoutSeconds: 30,
  });

  // Create a producer
    sendTimeoutMs: 30000,
    batchingEnabled: false,
    initialSequenceId: 5294967295,
  });

  console.log('initialSequenceId = ' + producer.getLastSequenceId());
  // Send messages
  for (let i = 0; i < 100; i += 1) {
    const msg = `my-message-${i}`;
    await producer.send({
      data: Buffer.from(msg),
    });
  }
  console.log('lastSequenceId = ' + producer.getLastSequenceId());

  await producer.close();
  await client.close();
})();
```

alanhoff:
@hrsakai @sijie sorry for the delay here, let's start by correcting some of my misconceptions: 

JS do support ints higher than 32 bits when representing them as a [`Number`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER), but it *may* start losing precision for ints above 2<sup>53</sup> - 1 (or lower than the negative equivalent). Example:

```javascript
const x = Number.MAX_SAFE_INTEGER + 1;
const y = Number.MAX_SAFE_INTEGER + 2;

console.log(x === y); // will output true
```

That said, we're "kinda" safe at the moment since it's the native addon that's [incrementing the sequence](https://github.com/apache/pulsar/blob/1cf29f244a783f62aa6e4b6860135758e0f8034c/pulsar-client-cpp/lib/ProducerImpl.cc#L65) using a proper `int64_t`, however IMHO I don't think the JS client should be using the `Number` primitive to represent an `int64_t` since that's unsafe.

The correct thing to do would be changing all sequences so they accept a [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) instead of a Number, but it would be a breaking change and it wouldn't support Node.js < `10.4.0`. So in the meanwhile this PR allows devs to optionally pass sequences as strings which are being casted into `int64_t` by the addon without losing precision.

hrsakai:
@alanhoff 
Thank you for your explanation.

Could you change initialSequenceId to accept NUMBER and BigInt and getLastSequenceId() to return BigInt?
We can accept that we don't support `Node.js < 10.4.0`.

Issue #94, opened 2020-06-11
Message ID for produced messages


psr323:
Hi,
Currently we are not getting any response for a produced message. Is there a way to get the message ID for a message that was produced? If not, can we please have that feature?

k2la:
There is not such a way at present but we will support it in the future.
It may take some time to support it as `pulsar-client-cpp` needs to be modified.

psr323:
Hi @wolfstudy is this feature available for use now?

massakam:
@psr323 No. `pulsar-client-cpp` has been modified, but it has not been released yet. The code in this repository will need to be modified after `pulsar-client-cpp` is released, so it will take some time before this feature is available.

PR #137, opened 2020-12-25
Return MessageId as promise resolve at Producer#send

equanz:
Fixes https://github.com/apache/pulsar-client-node/issues/94

### Motivation
We want to get MessageId at `Producer#send` method.

### Modifications
* Change Producer#send Promise response
  - from `Promise<null>` to `Promise<Object>`
  - return MessageId as `Promise<Object>`
* Change C++ library version to `2.7.0`
  - because of https://github.com/apache/pulsar/pull/7439

### Verifying this change
Add end-to-end test to test cases.


tests/end_to_end.test.js
```diff
@@ -665,5 +665,54 @@ const Pulsar = require('../index.js');
...
+        topic,
+        sendTimeoutMs: 30000,
+        batchingEnabled: true,
+      });
+      expect(producer).not.toBeNull();
+
+      const consumer = await client.subscribe({
+        topic,
+        subscription: 'sub1',
+        ackTimeoutMs: 10000,
```
massakam, line 686: Do we need to set `ackTimeoutMs`?

equanz: No, it isn't necessary. I'll remove it.

equanz:
PTAL

massakam (approved):

PR #145, opened 2021-03-09
Add license headers to some files

massakam:
Some files did not have license headers, so added them. In addition, I rolled back `license-check-and-add` to the previous version because `npm run license:addheader` didn't working properly.


nkurihar (approved):
LGTM

PR #146, opened 2021-03-09
Add definition of AthenzConfig

massakam:
The constructor of the `AuthenticationAthenz` class can accept one string or object as an argument, but currently there is only a type definition for string. So I added a type definition to accept the object type argument.


nkurihar (approved):
LGTM

PR #149, opened 2021-04-06
Support Disable Replication

k2la:
Support disable replication.

Because `disableReplication` is not working in pulsar-client-cpp 2.7.0 or previous version, needs 2.7.1 or later version. 
This bug was fixed in https://github.com/apache/pulsar/pull/9372


nkurihar (approved):
LGTM

PR #151, opened 2021-04-16
Add type definition for disableReplication property

massakam:
Added the type definition of the `disableReplication` property added in https://github.com/apache/pulsar-client-node/pull/149.


nkurihar (approved):
LGTM

PR #153, opened 2021-05-11
Support Reader Listener

k2la:
Add listener function to the Reader, using the C++ client listener support.
- Add listener for reader
- Add example and e2e test of using the reader listener

### DIFF
--- a/src/Reader.cc
+++ b/src/Reader.cc
@@ -108,11 +108,16 @@ class ReaderNewInstanceWorker : public Napi::AsyncWorker {
                                     this->readerConfig->GetCReaderConfig(), &this->readerWrapper->cReader);
     if (result != pulsar_result_Ok) {
       SetError(std::string("Failed to create reader: ") + pulsar_result_str(result));
+      if (this->readerConfig) {
+        delete this->readerConfig;
+      }
       return;
     } else {
       this->listener = this->readerConfig->GetListenerCallback();
     }
-    delete this->readerConfig;
+    if (this->readerConfig) {
+      delete this->readerConfig;
+    }
   }
   void OnOK() {
     Napi::Object obj = Reader::constructor.New({});
@@ -254,5 +259,4 @@ Reader::~Reader() {
   if (this->listener) {
     this->CleanupListener();
   }
-  pulsar_reader_free(this->wrapper->cReader);
 }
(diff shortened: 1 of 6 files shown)


src/Reader.cc
```diff
@@ -66,33 +105,41 @@ class ReaderNewInstanceWorker : public Napi::AsyncWorker {
...
-                                    this->readerConfig->GetCReaderConfig(), &(this->cReader));
-    delete this->readerConfig;
+                                    this->readerConfig->GetCReaderConfig(), &this->readerWrapper->cReader);
     if (result != pulsar_result_Ok) {
       SetError(std::string("Failed to create reader: ") + pulsar_result_str(result));
       return;
+    } else {
+      this->listener = this->readerConfig->GetListenerCallback();
     }
+    delete this->readerConfig;
```
hrsakai, line 115: if `result` is not ok, won't this line be executed?

src/Reader.cc
```diff
@@ -176,13 +228,31 @@ class ReaderCloseWorker : public Napi::AsyncWorker {
...
+  this->Unref();
+  this->listener->callback.Release();
+  this->listener = nullptr;
+}
+
+Reader::~Reader() {
+  if (this->listener) {
+    this->CleanupListener();
+  }
+  pulsar_reader_free(this->wrapper->cReader);
```
hrsakai, line 257: Can we delete this line because `pulsar_reader_free` is executed on `CReaderWrapper::~CReaderWrapper`?

k2la: OK. I removed it.

k2la: Exactly. If the result is not ok, this is not be executed.
I fixed that this is be done if the result is not ok.

src/Reader.cc
k2la: Exactly. If the result is not ok, this is not be executed.
I fixed that this is be done if the result is not ok.

hrsakai: I think it is better to write as follows.
https://github.com/apache/pulsar-client-node/blob/v1.3.0/src/Consumer.cc#L180-L187

k2la: I implemented `createReaderCallback` like [subscribeCallback](https://github.com/apache/pulsar-client-node/blob/v1.3.0/src/Consumer.cc#L178-L190).

hrsakai (approved):
LGTM

PR #156, opened 2021-05-19
Call client.close on tests

hrsakai:
This PR is to avoid segmentation fault on ci tests.
I don't know root cause, but it seems that we can avoid segmentation fault by adding client.close after test is finished.


k2la (approved):
LGTM

PR #163, opened 2021-07-20
Use Buffer::Copy instead of Buffer::New

hrsakai:
In our enviroment, sometimes we get following error.
```
#
# Fatal error in , line 0
# Check failed: result.second.
#
#
#
#FailureMessage Object: 0x7ffe7bd5eda0
 1: 0xa92851  [node]
 2: 0x1a1ac64 V8_Fatal(char const*, ...) [node]
 3: 0xe7e119 v8::internal::GlobalBackingStoreRegistry::Register(std::shared_ptr<v8::internal::BackingStore>) [node]
 4: 0xbc7e48 v8::ArrayBuffer::GetBackingStore() [node]
 5: 0x9e3a80 napi_get_typedarray_info [node]
 6: 0x7f815dd9f669 Message::GetData(Napi::CallbackInfo const&) [/root/pulsar-client-node/build/Release/Pulsar.node]
 7: 0x7f815dda3aa1 Napi::ObjectWrap<Message>::InstanceMethodCallbackWrapper(napi_env__*, napi_callback_info__*) [/root/pulsar-client-node/build/Release/Pulsar.node]
 8: 0x9da63f  [node]
 9: 0xc06acb  [node]
10: 0xc08076  [node]
11: 0xc086f6 v8::internal::Builtin_HandleApiCall(int, unsigned long*, v8::internal::Isolate*) [node]
12: 0x1427159  [node]
Illegal instruction (core dumped)

```

## Modification
I use `Napi::Buffer<T>::Copy` instead of `Napi::Buffer<T>::New`.

## Steps to reproduce
One easy way to reproduce  is to call `msg::getData()` twice.
```
    const msg = await consumer.receive();
    console.log(msg.getData().toString());
    console.log(msg.getData().toString());
```

<br>
This issue occures on node 14 or above.

### DIFF
--- a/src/Message.cc
+++ b/src/Message.cc
@@ -102,7 +102,7 @@ Napi::Value Message::GetData(const Napi::CallbackInfo &info) {
   }
   void *data = const_cast<void *>(pulsar_message_get_data(this->cMessage));
   size_t size = (size_t)pulsar_message_get_length(this->cMessage);
-  return Napi::Buffer<char>::New(env, (char *)data, size);
+  return Napi::Buffer<char>::Copy(env, (char *)data, size);
 }
 
 Napi::Value Message::GetMessageId(const Napi::CallbackInfo &info) {


massakam:
The following part doesn't need to be modified because `ptr` is not shared_ptr, right?
https://github.com/apache/pulsar-client-node/blob/07c6b1dd8132aed4e3d50ba182c25a7e8e3cf378/src/MessageId.cc#L100

hrsakai:
> The following part doesn't need to be modified because ptr is not shared_ptr, right?

This issue seems to occure if adding same pointer to `Napi::Buffer<T>::New`, whether shared_ptr or not.
For example, following code occures the error:
```
Napi::Value MessageId::Serialize(const Napi::CallbackInfo &info) {
  Napi::Env env = info.Env();

  int len;
  void *ptr = pulsar_message_id_serialize(GetCMessageId(), &len);

  // OK
  Napi::Buffer<char>::New(env, (char *)ptr, len, serializeFinalizeCallback);

  // Check failed: result.second.
  return Napi::Buffer<char>::New(env, (char *)ptr, len, serializeFinalizeCallback);
}

```


In fact, `pulsar_message_id_serialize` executes `memcpy` and returns the copy each time it is called, so same pointer is not added to `Napi::Buffer<T>::New` in `MessageId::Serialize` and so we don't need to modify `MessageId::Serialize`.
https://github.com/apache/pulsar/blob/6704f12104219611164aa2bb5bbdfc929613f1bf/pulsar-client-cpp/lib/c/c_MessageId.cc#L46-L53

massakam:
I got it. Thanks.

massakam (approved):
LGTM

PR #160, opened 2021-06-02
Fix leak at MessageId::Serialize

equanz:
### Motivation
`MessageId::Serialize` possibly have a memory leak bug. Here are a simple reproduce code and valgrind log.

```js
for (let i = 0; i < 1000; i += 1) {
  const msg = await consumer.receive();

  // serialize
  const serialize = msg.getMessageId().serialize();
  console.log(serialize);

  consumer.acknowledge(msg);
}
```

```
==1048== 7,000 bytes in 1,000 blocks are definitely lost in loss record 67 of 68
==1048==    at 0x50F4F73: malloc (vg_replace_malloc.c:309)
==1048==    by 0xA36C67A: pulsar_message_id_serialize (in /usr/lib/libpulsar.so.2.7.2)
==1048==    by 0x9F3E599: MessageId::Serialize(Napi::CallbackInfo const&) (in /app/node_modules/pulsar-client/build/Release/Pulsar.node)
==1048==    by 0x9F458D0: Napi::InstanceWrap<MessageId>::InstanceMethodCallbackWrapper(napi_env__*, napi_callback_info__*) (in /app/node_modules/pulsar-client/build/Release/Pulsar.node)
==1048==    by 0x9D7E5E: v8impl::(anonymous namespace)::FunctionCallbackWrapper::Invoke(v8::FunctionCallbackInfo<v8::Value> const&) (in /usr/bin/node)
==1048==    by 0xC03D0A: v8::internal::MaybeHandle<v8::internal::Object> v8::internal::(anonymous namespace)::HandleApiCallHelper<false>(v8::internal::Isolate*, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::FunctionTemplateInfo>, v8::internal::Handle<v8::internal::Object>, v8::internal::BuiltinArguments) (in /usr/bin/node)
==1048==    by 0xC052B5: v8::internal::Builtin_Impl_HandleApiCall(v8::internal::BuiltinArguments, v8::internal::Isolate*) (in /usr/bin/node)
==1048==    by 0xC05935: v8::internal::Builtin_HandleApiCall(int, unsigned long*, v8::internal::Isolate*) (in /usr/bin/node)
==1048==    by 0x1423358: Builtins_CEntry_Return1_DontSaveFPRegs_ArgvOnStack_BuiltinExit (in /usr/bin/node)
[...]

PR #159, opened 2021-05-31
Fix Leak in MessageId::ToString

k2la:
### Motivation
I run the following code with valgrind and it looks like there are a memory leak in MessageId::ToString.

```
  // Receive messages
  for (let i = 0; i < 1000; i += 1) {
    const msg = await consumer.receive();
    console.log(msg.getMessageId().toString());
    console.log(msg.getData().toString());
    consumer.acknowledge(msg);
  }
```

```
...
==20476== 13,890 bytes in 1,000 blocks are definitely lost in loss record 68 of 68
==20476==    at 0x4C29F73: malloc (vg_replace_malloc.c:309)
==20476==    by 0x5B03BD9: strndup (in /usr/lib64/libc-2.17.so)
==20476==    by 0x9EA3D74: pulsar_message_id_str (in /usr/lib/libpulsar.so.2.7.1)
==20476==    by 0x9A73E6A: MessageId::ToString(Napi::CallbackInfo const&) (in /home/k2la/pulsar-client-node/build/Release/Pulsar.node)
==20476==    by 0x9A7BFE0: Napi::ObjectWrap<MessageId>::InstanceMethodCallbackWrapper(napi_env__*, napi_callback_info__*) (in /home/k2la/pulsar-client-node/build/Release/Pulsar.node)
==20476==    by 0x9D1E04: v8impl::(anonymous namespace)::FunctionCallbackWrapper::Invoke(v8::FunctionCallbackInfo<v8::Value> const&) (in /usr/bin/node)
==20476==    by 0xC065F8: v8::internal::MaybeHandle<v8::internal::Object> v8::internal::(anonymous namespace)::HandleApiCallHelper<false>(v8::internal::Isolate*, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::FunctionTemplateInfo>, v8::internal::Handle<v8::internal::Object>, v8::internal::BuiltinArguments) (in /usr/bin/node)
==20476==    by 0xC083E6: v8::internal::Builtin_HandleApiCall(int, unsigned long*, v8::internal::Isolate*) (in /usr/bin/node)
==20476==    by 0x140DF98: ??? (in /usr/bin/node)
==20476==    by 0x1393723: ??? (in /usr/bin/node)
==20476==    by 0x13BDFD2: ??? (in /usr/bin/node)
==20476==    by 0x13E26D1: ??? (in /usr/bin/node)
...
```

### Modification
[...]

PR #168, opened 2021-08-06
Branch 1.3

Liapin:
Fixes
Use Buffer::Copy instead of Buffer::New #163
Fix leak at MessageId::Serialize #160
Fix Leak in MessageId::ToString #159

### DIFF
--- a/src/Message.cc
+++ b/src/Message.cc
@@ -100,7 +100,7 @@ Napi::Value Message::GetData(const Napi::CallbackInfo &info) {
   }
   void *data = const_cast<void *>(pulsar_message_get_data(this->cMessage));
   size_t size = (size_t)pulsar_message_get_length(this->cMessage);
-  return Napi::Buffer<char>::New(env, (char *)data, size);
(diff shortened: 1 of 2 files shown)


nkurihar:
@Liapin 
Hi, thank you for backporting these patches.
Though 1.4.0 has already been released, do you need 1.3.1 ?

Liapin:
Hi @nkurihar. Yes, it would be great to have 1.3.1

nkurihar:
@Liapin 
Thank you for your reply.

I guess the main reason for not 1.4.0 but 1.3.1 is the version of C++ Client, i.e., you need to upgrade the version from 2.7.0 to 2.8.0 in order to use 1.4.0.
Is my guess right?

Honestly speaking, we don't have any backporting plan for now...
Please give me some time to think about it.

nkurihar:
@Liapin
Could you rebase this PR?

I have cherry-picked the fix for segmentation fault on tests.
https://github.com/apache/pulsar-client-node/pull/170

Issue #181, opened 2021-11-27
A bug about MessageListenerProxy, got a nullptr of data->consumer(Consumer.cc#64)


lipeining:
os: Ubuntu 16.04.7 LTS
nodejs: v14.17.0
pulsar-client: 1.3.0
pulsar c++ client: 2.7.1
framework: midway.js

on our web server written with nodejs, when the server start, it will use pulsar-client(we clone a copy of branch-1.3) connect to our pulsar service and subscribe from some topics. but, with some bug, the server crash down after subscribe on a pulsar topic.would you please take sometime to find the crash reason, thanks a lot.

our web server subscribe on five or six topics, also, it had been work well done for a  long time. but, without change a line of js code, it crash down.
we spend some time to track what's happening.
when the node.js process crash down, we got the crash process core dump file,
here we got some log about why pulsar-client crash down.
### gcore
`gcore -c corefileName`
with help of gcore  and core dump file, we get some crash stack like this
```txt
thread #1: tid = 40071, 0x00007fffdfcd4af9 Pulsar.node`MessageListenerProxy(Napi::Env, Napi::Function, MessageListenerProxyData*) + 73, name = 'node', stop reason = signal SIGSEGV: invalid address (fault address: 0x10)
    frame #0: 0x00007fffdfcd4af9 Pulsar.node`MessageListenerProxy(Napi::Env, Napi::Function, MessageListenerProxyData*) + 73
  * frame #1: 0x00007fffdfcd5a37 Pulsar.node`std::_Function_handler<void (Napi::Env, Napi::Function), napi_status Napi::ThreadSafeFunction::BlockingCall<MessageListenerProxyData, void (*)(Napi::Env, Napi::Function, MessageListenerProxyData*)>(MessageListenerProxyData*, void (*)(Napi::Env, Napi::Function, MessageListenerProxyData*)) const::'lambda'(Napi::Env, Napi::Function)>::_M_invoke(std::_Any_data const&, Napi::Env&&, Napi::Function&&) + 23
    frame #2: 0x00007fffdfcbfbf2 Pulsar.node`Napi::ThreadSafeFunction::CallJS(napi_env__*, napi_value__*, void*, void*) + 130
    frame #3: 0x00000000009f3ab8 node`v8impl::(anonymous namespace)::ThreadSafeFunction::IdleCb(uv_idle_s*) + 328
    frame #4: node`uv__run_idle(loop=0x00000000044b0800) at loop-watcher.c:68
[...]

PR #182, opened 2021-11-29
fix: use pause and resume api to ensure orderly init listener->consumer

lipeining:


lipeining:
#181

massakam:
@lipeining Thank you for your contribution. The change looks good to me.

But why is the destination branch `branch-1.3` instead of `master`? Do you want v1.3.x with this patch to be released?

rickyes:
> @lipeining Thank you for your contribution. The change looks good to me.
> 
> But why is the destination branch `branch-1.3` instead of `master`? Do you want v1.3.x with this patch to be released?

Hi, @massakam 

We are currently experiencing this issue in version `v1.3.x` by chance and would like to know more details about the cause, if you know the details could you please help us analyze it ?

massakam:
@rickyes Details are explained in #181. Merging this PR should fix the issue.

massakam (approved):

rickyes:
@massakam It is rather strange that we do not reproduce it locally, but only in specific network environments.

massakam:
@rickyes This phenomenon does not always occur. It happens when `data->consumer` is accessed in the following part before it is initialized.
https://github.com/apache/pulsar-client-node/blob/50c090678ecedaa06e2c68ce1db3046655a9a075/src/Consumer.cc#L62-L65

Adding sleep to the following part will almost certainly reproduce this issue.
```diff
--- a/src/Consumer.cc
+++ b/src/Consumer.cc
@@ -23,6 +23,7 @@
 #include "MessageId.h"
 #include <pulsar/c/result.h>
 #include <atomic>
+#include <chrono>
 #include <thread>

 Napi::FunctionReference Consumer::constructor;
@@ -163,6 +164,7 @@ class ConsumerNewInstanceWorker : public Napi::AsyncWorker {
     Consumer *consumer = Consumer::Unwrap(obj);

     consumer->SetCConsumer(this->consumerWrapper);
+    std::this_thread::sleep_for(std::chrono::seconds(5));
     consumer->SetListenerCallback(this->listener);
     this->deferred.Resolve(obj);
   }
```

rickyes:
@massakam Is it possible that the MessageListener was triggered before we triggered the callback subscribeCallback?

massakam:
@rickyes I think this bug becomes apparent if the `MessageListener` is triggered before the `subscribeCallback` is completed.

rickyes:
@massakam Sorry, I meant before the `subscribeCallback` was triggered and not before the `subscribeCallback` was completed.

rickyes:
Because if it is possible to trigger the `MessageListener` before the `subscribeCallback` is triggered, this PR doesn't do much good.

massakam:
@rickyes Sorry, [this comment](https://github.com/apache/pulsar-client-node/pull/182#issuecomment-986538154) is incorrect. It seems that `MessageListener` is not triggered until the execution of `subscribeCallback` is completed.

`subscribeCallback` is executed on the following line in the C++ client library. And the C++ client will not start receiving messages until this method completes.
https://github.com/apache/pulsar/blob/a54c6c003c626cb16d90200ad81dd3ec37be2133/pulsar-client-cpp/lib/ClientConnection.cc#L877

rickyes:
@massakam LGTM, thank you for your patient answer~

PR #184, opened 2021-12-08
Add tarball files to .gitignore

massakam:
Recently, the release process for the Node.js client has been modified and added a step to create a tarball of this repository.
```sh
# Archive the source files
$ git archive HEAD --prefix=pulsar-client-node-1.X.0/ --output pulsar-client-node-1.X.0.tar.gz
```
I added two file extensions to `.gitignore` to prevent this tarball from being committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ build/
 node_modules/
 report/
 .idea
+*.tar.gz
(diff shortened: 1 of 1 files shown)


hrsakai (approved):
LGTM

PR #186, opened 2021-12-09
Update dependencies

massakam:
Updated the dependent packages to the latest version with the same major version. `node-pre-gyp` has already been deprecated, so replaced it with an alternative package `@mapbox/node-pre-gyp`


hrsakai (approved):
LGTM

PR #190, opened 2022-01-15
Feature support oauth2 for node client

tuteng:
* The cpp client already supports oauth2 authentication, and some users want to support this feature in the node js client as well, so enable it in the node js client


package-lock.json
```diff
@@ -1,9204 +1,8 @@
 {
   "name": "pulsar-client",
   "version": "1.6.0-rc.0",
-  "lockfileVersion": 2,
```
jiazhai: Seems we removed a lot of code, what were these codes used for?

jiazhai (approved):

gaoran10 (approved):
Lgtm

package-lock.json
tuteng: Thanks, revert this change

hassanzia32:
Is there a timeline for when the 1.6.0 version is scheduled to be released?

PR #194, opened 2022-02-15
Fixed windows install client

tuteng:
* Support for specified architectures

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,17 @@ If an incompatible version of the C++ client is installed, you may fail to build
 
 ### Install on windows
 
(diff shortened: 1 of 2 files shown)


BewareMyPower (commented):
Should we update the document as well? I think we need to define `OS_ARCH` environment variable when we compile the node client.

tuteng:
> Should we update the document as well? I think we need to define `OS_ARCH` environment variable when we compile the node client.

Yes I think the documentation needs to be updated, I also think os_arch is an optional field, if not defined it will use this `\vcpkg_installed\bin\` to find dependencies

gaoran10 (approved):

BewareMyPower (approved):

PR #215, opened 2022-05-09
prevent double callback segfault in client close

Matt-Esch:
There appears to be a race condition that causes the close callback to execute twice on close. I believe that a similar issue existed with the old synchronous api call in a async worker, it just materialised slightly differently. There is a proposed fix for it here https://github.com/apache/pulsar/pull/15508

This PR temporarily fixes the problem, though I think a better solution should be decided on. For example

  - I'm writing to `std::cerr`, I'm not convinced this is generally acceptable, but I cannot guarantee the lifetime of the logger
  - The behaviour for old pulsar clients should be slightly different to the behaviour once cpp client is fixed, so there probably needs to be a version check. We don't want older clients to segfault, so either we bump the min version supported or we need to actively check the pulsar client version before deciding how to handle the close callback.

### DIFF
--- a/src/Client.cc
+++ b/src/Client.cc
@@ -17,6 +17,7 @@
  * under the License.
  */
 
+#include <iostream>
 #include "Authentication.h"
 #include "Client.h"
 #include "Consumer.h"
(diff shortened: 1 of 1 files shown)


massakam:
If https://github.com/apache/pulsar/pull/15508 fundamentally solves this issue, I think it's enough to raise the minimum version of the C++ client required by the Node.js client by the next release. I don't think it's necessary to merge this workaround.

PR #235, opened 2022-10-14
[feat] Napi support package c++ static binary.

shibd:
### Motivation

Currently, users using the `node js client` must first install the `C++ client`. And when the user executes `npm install`, it also needs to be compiled. This is not user-friendly.

This PR is to provide pre-built packages for all the supported environments, using [GYP]() and rebuild. The pre-built package is then uploaded to [`Github Release`](https://github.com/shibd/pulsar-client-node/releases/tag/v1.8.1-rc.3), and when the user installs the client, the system version is matched and the pre-built package is downloaded.

Currently tested `linux_glibc_x86_64`, `linuc_musl_x86_64`, `linux_glibc_arm64`, `linuc_musl_arm64` and `mac_x86_64`. I've built it in my [fork repository](https://github.com/shibd/pulsar-client-node/releases/tag/v1.8.1-rc.3) and can install it directly using the following command (without first installing the C++ client).

```
npm install shibaodi-pulsar-client
```

#### New Github workflows

- When trigger release: 
   1. Build the binary C++ static library first(use new c++ client 3.0 source).
   2. And use [node-pre-gyp](https://github.com/mapbox/node-pre-gyp) pre-built `napi` packages.
   3. Use [node-pre-gyp-github](https://github.com/bchr02/node-pre-gyp-github) publish this package to `GitHub Release`.
- When trigger unit test(use ubuntu GitHub runner):
   1. Install c++ lib(use new c++ client 3.0 version).
   2. Use this c++ lib build `napi`.
   3. Use pulsar image start broker(use pulsar latest).
   5. Run the unit test.

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

### TODO
- [ ] Add pre-build windows binary package.
- [ ] Change user docs.


BewareMyPower (changes requested):
Why do you rename the directory name from `certificate` to `test-conf`?

examples/encryption-consumer.js
```diff
@@ -32,7 +32,7 @@ const Pulsar = require('pulsar-client');
     subscription: 'sub1',
     subscriptionType: 'Shared',
     ackTimeoutMs: 10000,
-    privateKeyPath: "./certificate/private-key.client-rsa.pem"
+    privateKeyPath: "./test-test-conf/private-key.client-rsa.pem"
```
BewareMyPower, line 35: This path is appears to be wrong. It should be `./tests/test-conf`.

BTW, you should not change JS code to adapt your test script. Instead, you should change your test script to adapt the existing code, which is a little different from Python client.

shibd:
> Why do you rename the directory name from `certificate` to `test-conf`?

Just for the unified, put all the test-related configurations together.

BewareMyPower:
> Just for the unified

There is no need to make two separate repositories consistent. What if the Python client repo renamed the directory or adjust the hierarchy later? **In this case, do you also want to sync the changes here?** People who are interested in Node.js client might not be interested in Python client. That's why do we separate C++/Python repos.

From my perspective, if you have a huge PR, please remove unnecessary changes as much as possible.

shibd:
> From my perspective, if you have a huge PR, please remove unnecessary changes as much as possible.

Make sense. I removed unnecessary changes. PTAL.

shibd:
Currently completed:
- linux_glibc_x86_64
- linux_musl_x86_64
- linux_glibc_arm64
- linux_musl_arm64
- macOS_x86_64

Please refer to my repo release: https://github.com/shibd/pulsar-client-node/releases/tag/v1.8.1-rc.3

Other environments:
- `macOS_arm64`: Not supported at the moment. Although the `cpp client` can be properly compiled on the `x86_64` by cross-compilation, it is not successful when packaging `NAPI`. Maybe we need a `macOS_m1` runner to run the compilation.
- `windows_x86_64`: I'm doing compilation testing, which will be added later.

@merlimat Do you feel there are other environments that need support?

@BewareMyPower @merlimat This PR can be reviewed first when you have time. Thanks.

.github/workflows/ci-build-release-napi.yml
```diff
@@ -0,0 +1,130 @@
...
+            NODE_VERSION=${{matrix.nodejs}}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+
+      - name: Build and Test NAPI file
+        run: |
+          docker run -i -v $PWD:/pulsar-client-node build:latest \
+              /pulsar-client-node/pkg/build-napi-inside-docker.sh
+
+      - name: Upload Node binaries lib to GitHub Release
```
shibd, line 124: Need to add a secrets to this repo.

merlimat: We could try to upload these as github actions artifacts. Then, we can automate to fetch them at release time and add them to the ASF dist Subversion, similar to what we do for C++/Python.  See: 

Then, we can just point GYP to the HTTP URL from ASF download.

.github/workflows/ci-build-release-napi.yml
```diff
@@ -0,0 +1,130 @@
+name: Build release napi
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
```
merlimat, line 6: We should probably not running this for `master`

.github/workflows/ci-pr-validation.yml
```diff
@@ -0,0 +1,40 @@
...
+    runs-on: ubuntu-22.04
+    timeout-minutes: 120
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@v3
+
+      - name: Test
+        run: |
+          ./run-unit-tests.sh
```
merlimat, line 40: We should also run the pre-gyp build as part of PR validation (to ensure that it keeps working and we don't find the issues only when we're attempting to do a release). 

We should run it for a subset of targets though: eg. linux-glibc-x64_86 & mac_x86_64

pkg/mac/build-cpp-deps-lib.sh
```diff
@@ -0,0 +1,167 @@
...
+
+source $(dirname $0)/common.sh
+
+ZLIB_VERSION=1.2.13
+OPENSSL_VERSION=1_1_1q
+BOOST_VERSION=1.79.0
+PROTOBUF_VERSION=3.20.0
+ZSTD_VERSION=1.5.2
+SNAPPY_VERSION=1.1.3
+CURL_VERSION=7.61.0
```
merlimat, line 29: we can try to use the same format as for cpp/python: https://github.com/apache/pulsar-client-cpp/blob/main/dependencies.yaml and sync the versions as of now.

pkg/mac/build-cpp-lib.sh
```diff
@@ -0,0 +1,51 @@
...
+
+source $(dirname $0)/common.sh
+
+PULSAR_DIR=${MAC_BUILD_DIR}/build-pulsar
+PULSAR_PREFIX=${PULSAR_DIR}/install
+mkdir -p $PULSAR_PREFIX
+cd $PULSAR_DIR
+
+# Pulsar
+curl -O -L https://dist.apache.org/repos/dist/dev/pulsar/pulsar-client-cpp-${PULSAR_CPP_VERSION}-candidate-2/apache-pulsar-client-cpp-${PULSAR_CPP_VERSION}.tar.gz
```
merlimat, line 29: Add TODO comment to switch to official releases URL

pulsar-client-cpp-version.txt
```diff
@@ -0,0 +1 @@
+3.0.0
```
merlimat, line 1: Where are we going to define the version of the wrapper?

src/pulsar-binding.js
```diff
@@ -0,0 +1,8 @@
+const path = require('path');
+const binary = require('@mapbox/node-pre-gyp');
+
+const bindingPath = binary.find(path.resolve(path.join(__dirname, '../package.json')));
```
merlimat, line 4: Would this work on Windows?

merlimat (commented):
Very nice to see such progress on the packaging!

merlimat:
> macOS_arm64: Not supported at the moment. Although the cpp client can be properly compiled on the x86_64 by cross-compilation, it is not successful when packaging NAPI. Maybe we need a macOS_m1 runner to run the compilation.

We can try to find a way to cross-compile. Right now there are no M1 runners in GH actions. 
Let this not be a blocker.

pulsar-client-cpp-version.txt
shibd: In pakcage.json -> version

src/pulsar-binding.js
shibd: I'm starting to package Windows and I'll test it

.github/workflows/ci-build-release-napi.yml
shibd: I modified it, but I didn't test the whole process. I don't have permission to commit to svn.

shibd:
> We can try to find a way to cross-compile. Right now there are no M1 runners in GH actions.
Let this not be a blocker.

Hi, @merlimat I can cross-compile `libpulsarwithdeps.a`, but I can't cross-compile `napi`. I tried using the command '--target_arch=arm64', but it doesn't work. You can refer to failed job: https://github.com/shibd/pulsar-client-node/actions/runs/3294556450/jobs/5432190252 

I raised an issue [here](https://github.com/mapbox/node-pre-gyp/issues/669) and haven't waited for a response yet.

I'll finish compiling Windows first and then continue working on it here.

If you have a good idea, please recommend it to me. Thanks.

build-support/pulsar-test-service-stop.sh
```diff
@@ -1,4 +1,6 @@
-#!/bin/bash
+
+
+#!/usr/bin/env bash
```
merlimat, line 3: I think there should be no empty lines

build-support/stage-release.sh
```diff
@@ -0,0 +1,44 @@
...
+mkdir -p $DEST_PATH
+
+cd $PULSAR_NODE_PATH
+
+build-support/download-release-artifacts.py $WORKFLOW_ID $DEST_PATH
+build-support/generate-source-archive.sh $DEST_PATH
+
+# Sign all files
+#cd $DEST_PATH
+#find . -type f | xargs $PULSAR_NODE_PATH/build-support/sign-files.sh
```
merlimat, line 44: This should be uncommented

merlimat (approved):
👍 Looks good! Let's target windows build in a new PR

.github/workflows/ci-build-release-napi.yml
```diff
@@ -0,0 +1,136 @@
...
+        id: cache-dependencies
+        uses: actions/cache@v3
+        with:
+          path: pkg/mac/build
+          key: ${{ runner.os }}-${{ matrix.arch }}-mac-${{ hashFiles('pkg/mac/build-cpp-deps-lib.sh') }}
+
+      - name: Add env vars
+        shell: bash
+        run: |
+          if [ "${{ matrix.target }}" = "x86_64" ]; then
```
merlimat, line 45: Variable name `matrix.target` does not exists so it will resolve to empty.

.github/workflows/ci-build-release-napi.yml
```diff
@@ -0,0 +1,136 @@
...
+      - name: Build CPP lib
+        if: steps.cache-pulsar.outputs.cache-hit != 'true'
+        run: |
+          export ARCH=${{ matrix.arch }}
+          pkg/mac/build-cpp-lib.sh
+
+      - name: Build Node binaries lib
+        run: |
+          npm install --ignore-scripts
+          npx node-pre-gyp configure --target_arch=${{ env.TARGET }}
```
merlimat, line 66: instead of using `evn.TARGET`, we can use `matrix.arch` directly here.

pulsar-client-cpp-version.txt
merlimat: 👍

PR #238, opened 2022-10-27
[feat] Support windows build napi of static link c++ lib.

shibd:
### Motivation

#235 

Support windows build napi of static link c++ lib(x64 and x86).

### Modifications
- Download the C++ static library directly instead of recompiling

**Note**: Since there is no static library release for C++ yet, Use a [temporary release](https://github.com/BewareMyPower/pulsar-client-cpp/releases/tag/v3.1.0-rc-20221026-windows). 

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`


shibd:
@merlimat @BewareMyPower PTAL.

.github/workflows/ci-build-release-napi.yml
```diff
@@ -125,3 +125,68 @@ jobs:
...
+        run: |
+          node pkg/load_test.js
+
+      - name: Package Node binaries lib
+        run: |
+          npx node-pre-gyp package --target_arch=${{ env.TARGET }}
+
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v3
+        if: matrix.nodejs == 16
```
BewareMyPower, line 189: Why do you upload artifcats only for Node.js 16?

shibd: Oh, While I was coding. Node.js 16 is the TLS version. Now its TLS changed to 18.

shibd: This does not mean that users can only use Node.js 16, because the version of [napi](https://nodejs.org/api/n-api.html#node-api-version-matrix) and the version of Node.js are controlled separately.

The version of NAPI we use is: 4

https://github.com/apache/pulsar-client-node/blob/10ef71a0d235955d91bf763d12513eca51948df0/binding.gyp#L37

Then you can compare. `NAPI 4` applies to Node.js > 10. Refer here: https://nodejs.org/api/n-api.html#node-api-version-matrix

BewareMyPower: I mean, why do you run the build CI with both Node.js 16 and 18 but only 16 is installed. See the matrix you added:

```yaml
      matrix:
        arch:
          - x64
          - x86
        nodejs:
          - 16
          - 18
```

BewareMyPower: But this is `ci-build-release-napi.yml`, which is used to build the artifacts, rather than validation. If you want to verify the source code can be compiled with different versions of Node.js, you should add the workflows into `ci-pr-validation.yml`.

BewareMyPower (changes requested):
image

You can see the PR validation only verified the case of x86 Windows and Node.js 16. The Node.js version compatibility checks should be performed when pushing a PR, not when pushing a tag.

.github/workflows/ci-build-release-napi.yml
shibd: I thought about it, based on the [compatibility guarantee of `NAPI` and `Node.js`](https://nodejs.org/api/n-api.html#node-api-version-matrix). We no longer need to verify the versions of both `NAPI` and `Node.js`. I removed the node 16 parameter

shibd:
> image
> 
> You can see the PR validation only verified the case of x86 Windows and Node.js 16. The Node.js version compatibility checks should be performed when pushing a PR, not when pushing a tag.



> image
> 
> You can see the PR validation only verified the case of x86 Windows and Node.js 16. The Node.js version compatibility checks should be performed when pushing a PR, not when pushing a tag.

Yes, I added more verification conditions.

BewareMyPower (commented):
Is there any way to verify the node module built with NPI of Node.js 18 can be used by Node.js 16?

BewareMyPower (approved):
LGTM

The API and ABI forward compatibility verification is not a blocker. But it seems that the ABI compatibility can be guaranteed according to [here](https://nodejs.org/api/n-api.html#implications-of-abi-stability)

> Although Node-API provides an ABI stability guarantee, other parts of Node.js do not, and any external libraries used from the addon may not.

shibd:
> Is there any way to verify the node module built with NPI of Node.js 18 can be used by Node.js 16?

I guess no verification is needed, We built using version `4` of NAPI.

https://github.com/apache/pulsar-client-node/blob/10ef71a0d235955d91bf763d12513eca51948df0/binding.gyp#L37

Then refer to this [table](https://nodejs.org/api/n-api.html#implications-of-abi-stability) to know that both 18 and 16 can be used.

RobertIndie (approved):

Demogorgon314 (approved):

PR #241, opened 2022-11-01
[feat] Add support for seek by message id

RobertIndie:
Add seek by message id method to the consumer.


shibd:
BTW, Can you add seek by timestamp?

RobertIndie:
> BTW, Can you add seek by timestamp?

Let's add it in another PR.

shibd (approved):

PR #245, opened 2022-11-10
Upload binaries napi to release.

shibd:
### Motivation

At the end of the release CI, upload the binaries to `GitHub Release`. Make it easy for users to view and download these pre-built binaries.

PS: This has nothing to do with Apache's release process

### Modifications
- Add upload-release-action on release ci end.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/.github/workflows/ci-build-release-napi.yml
+++ b/.github/workflows/ci-build-release-napi.yml
@@ -70,6 +70,15 @@ jobs:
           name: macos-${{matrix.nodejs}}-${{matrix.arch}}
           path: build/stage/*/*.tar.gz
 
+      - name: Upload binaries to release
+        uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: build/stage/*/*.tar.gz
+          tag: ${{ github.ref }}
+          overwrite: true
+          file_glob: true
+
   linux-napi:
     name: Build NAPI ${{matrix.image}} - Node ${{matrix.nodejs}} - ${{matrix.cpu.platform}}
     runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


BewareMyPower (commented):
Could you show an example workflow (maybe in your fork) to show the uploaded binaries?

BewareMyPower (commented):
Do we need to upload the artifacts to the release page?

shibd:
> Do we need to upload the artifacts to the release page?

Yes, They are the same.

image

BewareMyPower (changes requested):
Sorry I didn't express clearly. You try to upload these artifacts to the release page, but is there any document to explain how to use it? For example, I downloaded napi-linux-musl-x64.tar.gz and uncompress it to have a `binding/Pulsar.node` file:

```
binding/Pulsar.node: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, with debug_info, not stripped
```

But I don't know how to use it because it's only a dynamic library. (BTW it looks the build is a debug build, maybe we should change it to a release build)

BewareMyPower:
Another concern is from comments [here](https://github.com/apache/pulsar-client-cpp/pull/72#issuecomment-1293494580), @merlimat what's your opinion about this PR that uploads some artifacts into the release page?

shibd:
> You try to upload these artifacts to the release page, but is there any document to explain how to use it?

Oh, I want to clarify that this has no impact on the average user. They are used through the `npm install pulsar-client`.

But for users who don't want to use the npm management package. They can be downloaded binaries through the `release page`.

Then use it like below(BTW: This approach does not require additional instructions in the documentation): 
```node
const Pulsar = require('Pulsar.node');

(async () => {
  // Create a client
  const clientConfig = {
  };

  const client = new Pulsar.Client(clientConfig);
  await client.close();
})();


```

BewareMyPower:
> But for users who don't want to use the npm management package. 

Why not use the the NPM package since the artifacts uploaded to the release page are the same with the artifacts uploaded into NPM? If the artifacts are from a candidate version, it's better not to upload the artifacts to GitHub release page.

BewareMyPower:
> Then use it like below(BTW: This approach does not require additional instructions in the documentation):
> 
> ```js
> const Pulsar = require('./Pulsar.node');
> ```

It doesn't make sense to me. The examples under the https://github.com/apache/pulsar-client-node/tree/master/examples directory all uses:

```node
```

It a user switched to a temporary release from an official release, they have to modify the existing code.

shibd:
> > But for users who don't want to use the npm management package.
> 
> Why not use the the NPM package since the artifacts uploaded to the release page are the same with the artifacts uploaded into NPM? If the artifacts are from a candidate version, it's better not to upload the artifacts to GitHub release page.

Make sense. I closed this PR.

PR #246, opened 2022-11-16
[feat][api] Expose resume and pause listeners

jaysonsantos:
This should give developers the ability to limit the inflight messages with some sort of semaphore/tokenbucket

### Motivation

The idea here is that because of the async nature of node, you could endup with an infinite number of messages in flight if you use the listener feature to send a callback. Allowing the user to access the pause and resume listener, allows them to pause/resume based on the number of in-flight messages. 

### Verifying this change

I just reused the [current test](https://github.com/apache/pulsar-client-node/blob/996bd91d13746e22410f22083db3637864315f07/tests/end_to_end.test.js#L192-L193) to pause and resume the listener.

### Documentation

- [x] `doc-required` 

- [ ] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/tests/end_to_end.test.js
+++ b/tests/end_to_end.test.js
@@ -189,6 +189,8 @@ const Pulsar = require('../index.js');
           const data = message.getData().toString();
           results.push(data);
           messageConsumer.acknowledge(message);
+          messageConsumer.pauseMessageListener();
+          messageConsumer.resumeMessageListener();
           if (results.length === 10) finish();
         },
       });
--- a/index.d.ts
+++ b/index.d.ts
@@ -98,6 +98,8 @@ export class Consumer {
   negativeAcknowledgeId(messageId: MessageId): void;
   acknowledgeCumulative(message: Message): Promise<null>;
   acknowledgeCumulativeId(messageId: MessageId): Promise<null>;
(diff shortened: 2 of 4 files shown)


tests/end_to_end.test.js
```diff
@@ -189,6 +189,8 @@ const Pulsar = require('../index.js');
           const data = message.getData().toString();
           results.push(data);
           messageConsumer.acknowledge(message);
+          messageConsumer.pauseMessageListener();
```
shibd, line 192: Please add unit tests for both interface usage scenarios.

shibd (changes requested):

massakam:
@jaysonsantos 
Could you add a test case as @shibd commented? In addition, please resolve the conflict.

RobertIndie (changes requested):
I think this is an issue related to https://github.com/apache/pulsar-client-node/issues/308. It's been fixed by https://github.com/apache/pulsar-client-node/pull/309.

Note that we don't have `pauseMessageListener` and `resumeMessageListener` methdo for the Java client API. If we want to add new APIs that don't exist in the Java client. We need to raise a PIP to discuss it. Anyway, I don't think we need these two APIs.

PR #249, opened 2022-11-25
Fix build MacOS x64 binary name error.

shibd:
### Motivation

In MacOS x64, We pre-order build binary filenames are: napi-darwin-unknown-`x86_64`.tar.gz

However, the arch obtained by [Node is](https://nodejs.org/api/os.html#osarch) x64. This way, when the user executes `npm install`, the binary file will not be found. Because it will go to request a download: napi-darwin-unknown-`x64`.tar.gz

### Modifications

- Change arch to `x64`.

### Documentation

- [x] `doc-required` 

- [ ] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/.github/workflows/ci-build-release-napi.yml
+++ b/.github/workflows/ci-build-release-napi.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         arch:
-          - x86_64
+          - x64
 #          - arm64
         nodejs:
           - 18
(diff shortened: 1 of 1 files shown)


BewareMyPower (approved):

RobertIndie (approved):

.github/workflows/ci-build-release-napi.yml
```diff
@@ -41,13 +41,16 @@ jobs:
       - name: Build CPP dependencies lib
         if: steps.cache-dependencies.outputs.cache-hit != 'true'
         run: |
-          export ARCH=${{ matrix.arch }}
+          if [ "${{ matrix.arch }}" = "x64" ]; then
+            export ARCH=x86_64
+          else
+            export ARCH=${{ matrix.arch }}
+          fi
```
tisonkun, line 48: It confuses me that we changed `arch` to `x64` above be resolved it to `x86_64` again here - what's the difference?

shibd: Because in `build-cpp-deps-lib.sh`, we use `x86_64` parameters uniformly. Used for configuration `CFLAGS` or `CXXFLAGS`

https://github.com/apache/pulsar-client-node/blob/10ef71a0d235955d91bf763d12513eca51948df0/pkg/mac/build-cpp-deps-lib.sh#L150-L151

shibd:
Close this, as #250 already contains it.

PR #250, opened 2022-11-27
Supoort prebuild macos-arm64 arch binaries.

shibd:
### Motivation

#235 and #238 support prebuild binaries. However, the MacOS arm64 arch is not yet supported.

### Modifications
- Supoort prebuild macos-arm64 arch binaries.
- Fixed the x64 architecture prebuild name error. #249  

### Verifying this change
You can download it [here](https://github.com/shibd/pulsar-client-node/suites/9522179570/artifacts/452316640) to validate the ARM64 prebuild binaries.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/.github/workflows/ci-build-release-napi.yml
+++ b/.github/workflows/ci-build-release-napi.yml
@@ -19,8 +19,8 @@ jobs:
       fail-fast: false
       matrix:
         arch:
-          - x86_64
-#          - arm64
+          - x64
+          - arm64
         nodejs:
           - 18
     steps:
@@ -38,16 +38,24 @@ jobs:
           path: pkg/mac/build
           key: ${{ runner.os }}-${{ matrix.arch }}-mac-${{ hashFiles('pkg/mac/build-cpp-deps-lib.sh') }}
 
+      - name: Add arch env vars
+        run: |
+          if [ "${{ matrix.arch }}" = "x64" ]; then
+            echo "ARCH=x86_64" >> $GITHUB_ENV
+          else
+            echo "ARCH=${{ matrix.target }}" >> $GITHUB_ENV
+          fi
+
       - name: Build CPP dependencies lib
         if: steps.cache-dependencies.outputs.cache-hit != 'true'
         run: |
-          export ARCH=${{ matrix.arch }}
+          export ARCH=${{ env.ARCH }}
           pkg/mac/build-cpp-deps-lib.sh
 
       - name: Build CPP lib
         if: steps.cache-pulsar.outputs.cache-hit != 'true'
         run: |
-          export ARCH=${{ matrix.arch }}
+          export ARCH=${{ env.ARCH }}
           pkg/mac/build-cpp-lib.sh
 
       - name: Build Node binaries lib
@@ -57,6 +65,7 @@ jobs:
           npx node-pre-gyp build --target_arch=${{ matrix.arch }}
 
       - name: Test loading Node binaries lib
+        if: matrix.arch == 'x64'
         run: |
           node pkg/load_test.js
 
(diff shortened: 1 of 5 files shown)


dependencies.yaml
```diff
@@ -24,5 +24,5 @@ protobuf: 3.20.0
 zlib: 1.2.13
 zstd: 1.5.2
 snappy: 1.1.9
-openssl: 1.1.1q
+openssl: 1.1.1s
```
BewareMyPower, line 27: Why did you upgrade the OpenSSL version? It's okay to add a patch to an existing OpenSSL release like the line removed in this PR.

```bash
        echo -e "#include <string.h>\n$(cat test/v3ext.c)" > test/v3ext.c
```

It's better to use a separate PR to upgrade the dependency version.

shibd: This is an OpenSSL bug, It was fixed in 1.1.1s, I fixed it by the way.

https://github.com/openssl/openssl/issues/18720#issuecomment-1327885780

BewareMyPower: It's just a compilation bug. The previous scripts already fixes it by adding the patch to include the `string.h`.

BewareMyPower: There is also a bug when building Python 3.7 from source, adding a patch is a common used fixed method. https://github.com/apache/pulsar-client-python/blob/c5c177af647997d1182f3d4a69c74824f39af38f/pkg/mac/build-dependencies.sh#L73

Though it's okay to upgrade the OpenSSL dependency, but it's not related to this PR.

BewareMyPower (approved):

dependencies.yaml
shibd: Okay, I submit in #251.

.github/workflows/ci-build-release-napi.yml
```diff
@@ -57,6 +65,7 @@ jobs:
           npx node-pre-gyp build --target_arch=${{ matrix.arch }}
 
       - name: Test loading Node binaries lib
+        if: matrix.arch == 'x64'
```
RobertIndie, line 68: Why do you skip `arm64` here? I saw that it's empty from CI: https://github.com/apache/pulsar-client-node/actions/runs/3561255235/jobs/5981995900#step:5:5

.github/workflows/ci-build-release-napi.yml
```diff
@@ -38,16 +38,24 @@ jobs:
           path: pkg/mac/build
           key: ${{ runner.os }}-${{ matrix.arch }}-mac-${{ hashFiles('pkg/mac/build-cpp-deps-lib.sh') }}
 
+      - name: Add arch env vars
+        run: |
+          if [ "${{ matrix.arch }}" = "x64" ]; then
+            echo "ARCH=x86_64" >> $GITHUB_ENV
+          else
+            echo "ARCH=${{ matrix.target }}" >> $GITHUB_ENV
```
RobertIndie, line 46: What is `matrix.target` referred to here?

shibd: Oh, Wrong written here. 

As for why CI can pass, it is because the default value of arch in the `common.sh` is arm64

.github/workflows/ci-build-release-napi.yml
shibd: Because GitHub's runner arch is x64, it can't run Arm64 binaries

RobertIndie: Get it. Thanks.

RobertIndie (approved):

merlimat (approved):
👍 Nice!

PR #200, opened 2022-03-04
convert worker threads to async functions

Matt-Esch:
I have replaced the use of async workers with direct async functions by utilising a thread safe deferred (promise). The need for this comes from the fact that with workers, calling producer.send in order results in non-deterministic ordering, and awaiting every send would diminish the utility of having a batched mode. I also found that closing the client while worker threads were ready to execute resulted in segfaults, so this change ensures that most operations execute in the order they are called on the js side.

I've also found several places lacking in good memory management discipline and have patched a few leaks. I've opted to use shared pointers with custom destructors to wrap all of the returned pointers, referring the pulsar library code to check who owns these pointers. (I did not change Auth but this could also benefit). I have avoided excessively adding std::move manually as modern compilers are quite good at optimising this themselves, but I would welcome suggestions where appropriate.

I've changed the logger so that it is set globally. Currently the logger is supplied by the client configuration, but this gives the impression that the logger can be set per client. This is not true and the logger can only be set once for the lifetime of the process. The logger change commandeers the logger by default, avoiding unexpected logging to stdout and allows for the js log function to be switched out at runtime.

I have replaced most of the workers with async functions, the only ones I haven't replaced are the ones for which async functions are not exposed to the c api. I've raised an issue for this https://github.com/apache/pulsar/issues/14452, but it should be noted that the library is still prone to segfaulting if the read workers execute after the client close call.

This is my first time contributing to this library and I'm not particularly experienced with C++, but I have written node addons before. I have run the tests and checked that the linting and formatter have been run. If there are concerns around the style (such as the use of auto) I would welcome the feedback.

PR #252, opened 2022-12-01
[improve][client] Use async C function in the Consumer#receive method

equanz:
### Motivation

Related to: https://github.com/apache/pulsar-client-node/pull/200

We can't implement the receive method by async function in https://github.com/apache/pulsar-client-node/pull/200 because the cpp library didn't expose the async receive C function.
The async receive C function was implemented in https://github.com/apache/pulsar/pull/17452 . It was released in v3.0.0. So, we want to use it.

### Modifications

* Use `pulsar_consumer_receive_async` on Consumer#receive method
  - ~Use `PULSAR_VERSION` to provide backward compatibility between other cpp clients~

### Verifying this change

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
Because this PR doesn't introduce new features.

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/src/Consumer.cc
+++ b/src/Consumer.cc
@@ -247,16 +247,35 @@ class ConsumerReceiveWorker : public Napi::AsyncWorker {
 };
 
 Napi::Value Consumer::Receive(const Napi::CallbackInfo &info) {
-  Napi::Promise::Deferred deferred = Napi::Promise::Deferred::New(info.Env());
   if (info[0].IsUndefined()) {
-    ConsumerReceiveWorker *wk = new ConsumerReceiveWorker(deferred, this->cConsumer);
-    wk->Queue();
+    auto deferred = ThreadSafeDeferred::New(Env());
+    auto ctx = new ExtDeferredContext(deferred);
+    pulsar_consumer_receive_async(
+        this->cConsumer.get(),
+        [](pulsar_result result, pulsar_message_t *rawMessage, void *ctx) {
+          auto deferredContext = static_cast<ExtDeferredContext *>(ctx);
+          auto deferred = deferredContext->deferred;
+          delete deferredContext;
+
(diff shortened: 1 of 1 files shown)


src/Consumer.cc
```diff
@@ -247,16 +248,43 @@ class ConsumerReceiveWorker : public Napi::AsyncWorker {
 };
 
 Napi::Value Consumer::Receive(const Napi::CallbackInfo &info) {
-  Napi::Promise::Deferred deferred = Napi::Promise::Deferred::New(info.Env());
   if (info[0].IsUndefined()) {
+#if PULSAR_VERSION >= 3000000
```
RobertIndie, line 252: I don't think we need to handle the case of the pulsar CPP version before 3.0. For the next release, we will compile the node client based on the latest c++ client. If the user wants to compile the c++ client themselves, they should also use the latest c++ client. Otherwise, there will be a lot of redundant code for compatibility.

RobertIndie (changes requested):

src/Consumer.cc
shibd: +1

equanz: I got it. I modified this PR to use the new C function directly.
PTAL

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "pulsar-client",
-  "version": "1.8.1-rc.0",
+  "version": "1.8.0-rc.0",
```
shibd, line 3: Can you rebase master? In the master, the version was originally 1.8.0-rc.0

https://github.com/apache/pulsar-client-node/blob/5173415d03e774afceb3d89bd90c2f20bc5a0f0e/package-lock.json#L2-L3

shibd (approved):

PR #258, opened 2022-12-07
Remove useless dependencies.

shibd:
### Motivation

In #235, wrong introduction of `verbose` dependencies. It will cause the installation to fail using `yarn install`.

### Modifications
- Remove `verbose` dependence.
- Remove `node-pre-gyp-github`, is also useless dependence. It is responsible for publishing NAPI to Release page. Refer #245; We don't need to publish.
- Compile NAPI of macOS with yarn to verify that `yarn` is available.
- Update package.lock.json file.

### Verifying this change

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/.github/workflows/ci-pr-validation.yml
+++ b/.github/workflows/ci-pr-validation.yml
@@ -32,12 +32,31 @@ jobs:
     timeout-minutes: 120
 
     steps:
-      - name: checkout
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v3
+      - name: Use Node.js 18
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
 
       - name: Run Test
         run: |
           ./tests/run-unit-tests.sh
+
+  validation-yarn:
+    name: Validation use yarn install
+    runs-on: ubuntu-22.04
+    timeout-minutes: 120
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Use Node.js 18
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+
+      - name: Use yarn install
+        run: |
+          yarn install --ignore-scripts
 
   macos-napi:
     name: Build NAPI macos - Node ${{matrix.nodejs}} - ${{matrix.arch}}
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,8 +12,7 @@
       "dependencies": {
         "@mapbox/node-pre-gyp": "^1.0.9",
         "bindings": "^1.5.0",
-        "node-addon-api": "^4.3.0",
-        "verbose": "^0.2.3"
+        "node-addon-api": "^4.3.0"
       },
       "devDependencies": {
         "@seadub/clang-format-lint": "0.0.2",
@@ -32,7 +31,6 @@
         "jest": "^27.2.4",
         "license-check-and-add": "^2.3.6",
         "lodash": "^4.17.21",
-        "node-pre-gyp-github": "1.4.4",
         "typescript": "^4.1.3"
       },
       "engines": {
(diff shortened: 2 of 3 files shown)


.github/workflows/ci-pr-validation.yml
```diff
@@ -89,9 +89,9 @@ jobs:
 
       - name: Build Node binaries lib
         run: |
-          npm install --ignore-scripts
-          npx node-pre-gyp configure --target_arch=${{ matrix.arch }}
-          npx node-pre-gyp build --target_arch=${{ matrix.arch }}
+          yarn install --ignore-scripts
```
RobertIndie, line 92: We need to verify using both yarn and npm

shibd: fixed.

.github/workflows/ci-pr-validation.yml
```diff
@@ -90,8 +90,8 @@ jobs:
       - name: Build Node binaries lib
         run: |
           yarn install --ignore-scripts
```
BewareMyPower, line 92: Could you verify both `yarn install` and `npm install`?

```bash
echo "Verifying yarn install"
yarn install --ignore-scripts
# Maybe removing the `node_modules` directory is unnecessary
rm -rf node_modules/
echo "Verifying npm install"
npm install --ignore-scripts
```

tisonkun: Since we only cache for `npm` above:

```
cache: 'npm'
```

I don't think we should update the pkg manager here. Any motivation to make this change? If so, it can hide some subtle bugs.

shibd: > Could you verify both yarn install and npm install?

In fact, used `npm install` when building Windows as well as Linux Arch. This has verified that both yarn and npm can be used.

.github/workflows/ci-pr-validation.yml
shibd: In fact, used npm install when building Windows as well as Linux Arch. This has verified that both yarn and npm can be used.

I'll keep using `yarn` when compiling macOS.

shibd:
@BewareMyPower @RobertIndie @tisonkun I add `yarn.lock` to make sure developers who compile with yarn using the same dependent version.

.github/workflows/ci-pr-validation.yml
BewareMyPower: > Any motivation to make this change? 

@tisonkun Because there is a weird issue that `yarn install` failed for the current master while `npm install` succeeded.

shibd:
> Not sure whether should we add two lock files for both `npm` and `yarn`. When I ran `yarn install` in my local env, there was a warning:
> 
> ```shell
> $ yarn install --no-scripts
> yarn install v1.22.19
> warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.
> ```
> 
> After the `yarn install` command is done, the `yarn.lock` file changed a lot:
> 
> ```shell
> $ git diff yarn.lock | wc -l
> 7082
> ```

You are right. For developers, we should only use one way to manage dependencies.   I remove `yarn.lock`.

BTW: For users, It can optionally use `yarn` or `npm`. So, currently, we are not uploading package.lock.json to the npm repository.

In this PR, we verified the installation method using yarn.

.github/workflows/ci-pr-validation.yml
```diff
@@ -58,7 +58,6 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.nodejs }}
-          cache: 'npm'
```
shibd, line 61: @tisonkun I didn't change to using yarn cache because using yarn cache requires a yarn.lock file. Referring to this https://github.com/apache/pulsar-client-node/pull/258#issuecomment-1342579551, we don't add yarn.lock.

For this compilation step, I chose not to use a cache.

tisonkun: If we choose to check in the npm lock file, it's strange that we run CI tasks with `yarn`.

What if moving the pkg manager to `yarn` totally?

BewareMyPower: > What if moving the pkg manager to yarn totally?

I think we'd better not. The motivation is to prevent the regression happened again. I think we can add a separated workflow to verify the `yarn install` in a single env like Linux.

shibd: > I think we'd better not. The motivation is to prevent the regression happened again. I think we can add a separated workflow to verify the yarn install in a single env like Linux.

Reuse the current workflow to save CI execution time. `yarn install` command can take tens of seconds.

BewareMyPower: The workflow is executed concurrently. And it would not take too much time. See the x64 build for Node 18 on Linux, it only takes 5 minutes in total.

shibd: Ok, Changed.

shibd:
@tisonkun @RobertIndie @BewareMyPower PTAL.

tisonkun (approved):

RobertIndie (commented):
What's the motivation for removing `package-lock.json`?

shibd:
> What's the motivation for removing `package-lock.json`?

Not deleted. Only updated.

RobertIndie (approved):

BewareMyPower (approved):

PR #262, opened 2022-12-22
Add license headers

nkurihar:
Added license headers to some files by executing `npm run license:addheader`.


tisonkun (approved):
LGTM. Although I think we don't have to add license header to markdown files.

nkurihar:
> I think we don't have to add license header to markdown files.

I feel so too, but I'm not sure whether the license is necessarry even for .md files, so I have added same as https://raw.githubusercontent.com/apache/pulsar/master/README.md .

tisonkun:
@nkurihar Anyway, it does no harm :)

Merging...

PR #280, opened 2023-01-05
[refactor] optimize ci and build from source code.

shibd:
### Motivation

When building by source code on a Linux system,  If the CPP client is present locally, the installation CPP client may fail.

### Modifications

- Install the CPP client to a tmp path, not a system path.  In this way, even if the CPP client exists locally, it does not affect each other.
- Add README to explain that build from source code requires Node.js version greater than 16.18.
- Optimize file structure.

### Verifying this change

- In my repo verifying release workflows: https://github.com/shibd/pulsar-client-node/actions/runs/3845833859

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`


RobertIndie (commented):
Thanks for your PR. Overall looks good to me.

But is there any benefit to change the dockerfiles structure? The `Dockerfile` is the default file name of the docker. I think if there are no other problems, we should not change it only for other naming convention.

shibd:
The main thing is to put Linux-related builds files into a folder so that it is more uniform.

Similarly, the build result is referenced to the same folder.

binding.gyp:
```
        ['OS=="linux"', {
          "include_dirs": [
             "pkg/linux/pulsar-cpp/include"
          ],
          "libraries": [
             "../pkg/linux/pulsar-cpp/lib/libpulsarwithdeps.a"
          ],
        }]
```

RobertIndie (approved):

Issue #286, opened 2023-01-16
[Feature Request] Support CentOS 7


BewareMyPower:
### Motivation

The EOL of CentOS 7 is 2024-06-30, see https://wiki.centos.org/About/Product. However, the Node.js client does not support CentOS 7.

### Reproduce

First, build the `napi-xxx.tar.gz` for x86_64 glibc linux and run the `centos:7` container.

```bash
docker build --build-arg PLATFORM=x86_64 --build-arg ARCH=x86_64 --build-arg NODE_VERSION=18 \
  -t node-build-x86_64 -f pkg/linux/Dockerfile_linux_glibc .
docker run -i -v $PWD:/pulsar-client-node node-build-x86_64 \
  /pulsar-client-node/pkg/linux/build-napi-inside-docker.sh
docker run -v $PWD:/pulsar-client-node -it centos:7 /bin/bash
```

Then, run the following commands inside the container:

```bash
yum update -y
curl -O -L https://nodejs.org/download/release/v16.19.0/node-v16.19.0-linux-x64.tar.gz
tar zxf node-v16.19.0-linux-x64.tar.gz
export PATH=$PWD/node-v16.19.0-linux-x64/bin:$PATH
tar zxf /pulsar-client-node/build/stage/pulsar-client-node-1.9.0-rc.0/napi-linux-glibc-x64.tar.gz
```

Now, we can see `Pulsar.node` can be loaded because of the high `GLIBCXX` version for CentOS 7.

```bash
# node
Welcome to Node.js v16.19.0.
Type ".help" for more information.
> const Pulsar = require('./binding/Pulsar.node')
Uncaught:
Error: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by /root/binding/Pulsar.node)
    at Object.Module._extensions..node (node:internal/modules/cjs/loader:1249:18)
    at Module.load (node:internal/modules/cjs/loader:1043:32)
    at Function.Module._load (node:internal/modules/cjs/loader:878:12)
    at Module.require (node:internal/modules/cjs/loader:1067:19)
    at require (node:internal/modules/cjs/helpers:103:18) {
  code: 'ERR_DLOPEN_FAILED'
}
```

We can see the supported GLIBCXX versions of `libstdc++.so` on CentOS 7 are:

```bash
# strings /usr/lib64/libstdc++.so.6 | grep ^GLIBCXX_[0-9]
GLIBCXX_3.4
GLIBCXX_3.4.1
[...]

PR #288, opened 2023-01-17
Make it compatible with GLIBCXX_3.4.19 (CentOS 7)

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-node/issues/286

### Motivation

The maximum GLIBCXX version on CentOS 7 is 3.4.19. If we want to use the Pulsar Node client on CentOS 7, we have to upgrade the libstdc++.so. The reason is the libstdc++ version is too high for some systems. https://github.com/apache/pulsar-client-node/pull/285 downgrades the base image to `node:18-buster` but it still requires the `GLIBCXX_3.4.22`.

### Modifications
- Use `centos:7` as the base image to build `Pulsar.node` for glibc-based Linux distributions.
- Add `load_test.sh` to verify `Pulsar.node` can be loaded by other docker images and test Node.js 16 and 19 on Debian buster and bullseye.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/.github/workflows/ci-pr-validation.yml
+++ b/.github/workflows/ci-pr-validation.yml
@@ -167,6 +167,14 @@ jobs:
           docker run -i -v $PWD:/pulsar-client-node build:latest \
               /pulsar-client-node/pkg/linux/build-napi-inside-docker.sh
 
+      - name: Test NAPI file in other containers
+        if: matrix.image == 'linux_glibc'
+        run: |
+          ./tests/load-test.sh node:16-buster ${{matrix.cpu.platform}}
+          ./tests/load-test.sh node:16-bullseye ${{matrix.cpu.platform}}
+          ./tests/load-test.sh node:19-buster ${{matrix.cpu.platform}}
+          ./tests/load-test.sh node:19-bullseye ${{matrix.cpu.platform}}
+
   windows-napi:
     name: Build NAPI windows - Node ${{matrix.nodejs}} - ${{matrix.arch}}
     runs-on: windows-2022
--- a/pkg/linux/Dockerfile_linux_glibc
+++ b/pkg/linux/Dockerfile_linux_glibc
@@ -17,15 +17,19 @@
 # under the License.
 #
 
-ARG NODE_VERSION
+FROM centos:7
 
-FROM node:${NODE_VERSION}-buster
+RUN yum update -y
+RUN yum install -y gcc gcc-c++ make python3
 
-RUN apt-get update -y && \
-     apt-get install -y \
-        curl \
-        g++ \
-        make \
-        python3
+WORKDIR /app
+ARG PLATFORM
+RUN SUFFIX=$(echo ${PLATFORM} | sed 's/86_//') && \
+  echo $SUFFIX && \
+  curl -O -L https://nodejs.org/download/release/v16.19.0/node-v16.19.0-linux-$SUFFIX.tar.gz && \
+  tar zxf node-v16.19.0-linux-$SUFFIX.tar.gz && \
+  mv node-v16.19.0-linux-$SUFFIX node-v16.19.0
+
+ENV PATH="/app/node-v16.19.0/bin:$PATH"
 
 CMD ["sh"]
--- a/tests/docker-load-test.sh
+++ b/tests/docker-load-test.sh
@@ -25,11 +25,11 @@ set -ex
 # Create an empty directory to test
 mkdir -p /app && cd /app
 tar zxf /pulsar-client-node/pulsar-client-node.tar.gz
-cp /pulsar-client-node/build/Release/Pulsar.node .
 
-# Import the Pulsar.node instead of the node-gyp binding
-sed -i 's/src\/pulsar-binding/Pulsar/' index.js
-find src/ -name "*.js" | xargs sed -i 's/\.\/pulsar-binding/..\/Pulsar/'
+# Use the existing Pulsar.node built in a specific container
+mkdir -p lib/binding
+cp /pulsar-client-node/build/Release/Pulsar.node lib/binding/
+npm install
 
 # Test if Pulsar.node can be loaded
 node pkg/load_test.js
--- a/tests/docker-load-test.sh
+++ b/tests/docker-load-test.sh
@@ -28,7 +28,7 @@ tar zxf /pulsar-client-node/pulsar-client-node.tar.gz
 
 # Use the existing Pulsar.node built in a specific container
 mkdir -p lib/binding
-cp /pulsar-client-node/build/Release/Pulsar.node lib/binding/
+cp /pulsar-client-node/build/Release/pulsar.node lib/binding/
 npm install
 
 # Test if Pulsar.node can be loaded
--- a/tests/docker-load-test.sh
+++ b/tests/docker-load-test.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 #
 # Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file
@@ -17,12 +18,18 @@
 # under the License.
 #
 
-build/
-node_modules/
-report/
-.idea
-.vscode
-*.tar.gz
-*.tgz
-pkg/linux/pulsar-cpp/
-lib/
+# NOTE: This script should only run inside a Node.js docker container.
+# See ./load-test.sh for details.
+set -ex
+
+# Create an empty directory to test
+mkdir -p /app && cd /app
+tar zxf /pulsar-client-node/pulsar-client-node.tar.gz
+cp /pulsar-client-node/build/Release/Pulsar.node .
+
+# Import the Pulsar.node instead of the node-gyp binding
+sed -i 's/src\/pulsar-binding/Pulsar/' index.js
+find src/ -name "*.js" | xargs sed -i 's/\.\/pulsar-binding/..\/Pulsar/'
+
+# Test if Pulsar.node can be loaded
+node pkg/load_test.js
--- a/tests/load-test.sh
+++ b/tests/load-test.sh
@@ -35,9 +35,9 @@ PLATFORM=$2
 ROOT_DIR=${ROOT_DIR:-$(git rev-parse --show-toplevel)}
 cd $ROOT_DIR
 
-if [[ ! -f $ROOT_DIR/pulsar-client-node.tar.gz ]]; then
-    git archive -o pulsar-client-node.tar.gz HEAD
-fi
+git archive -o pulsar-client-node.tar.gz HEAD
 
 docker run --platform $PLATFORM -v $PWD:/pulsar-client-node $IMAGE \
     /bin/bash /pulsar-client-node/tests/docker-load-test.sh
+
+rm pulsar-client-node.tar.gz
--- a/tests/load-test.sh
+++ b/tests/load-test.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 #
 # Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file
@@ -17,12 +18,26 @@
 # under the License.
 #
 
-build/
-node_modules/
-report/
-.idea
-.vscode
-*.tar.gz
-*.tgz
-pkg/linux/pulsar-cpp/
-lib/
+# Test if the Pulsar.node built from ../pkg/linux/build-napi-inside-docker.sh
+# can be loaded in other Linux distributions.
+
+set -e
+
+if [[ $# -lt 2 ]]; then
+    echo "Usage $0 <node-image-name> <platform>
+
+See https://hub.docker.com/_/node for valid images"
+    exit 1
+fi
+IMAGE=$1
+PLATFORM=$2
+
+ROOT_DIR=${ROOT_DIR:-$(git rev-parse --show-toplevel)}
+cd $ROOT_DIR
+
+if [[ ! -f $ROOT_DIR/pulsar-client-node.tar.gz ]]; then
+    git archive -o pulsar-client-node.tar.gz HEAD
+fi
+
+docker run --platform $PLATFORM -v $PWD:/pulsar-client-node $IMAGE \
+    /bin/bash /pulsar-client-node/tests/docker-load-test.sh
--- a/pkg/linux/download-cpp-client.sh
+++ b/pkg/linux/download-cpp-client.sh
@@ -36,7 +36,9 @@ else
   PLATFORM=x86_64
 fi
 
+rm -rf $ROOT_DIR/pkg/linux/pulsar-cpp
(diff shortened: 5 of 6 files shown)


tests/docker-load-test.sh
```diff
@@ -0,0 +1,35 @@
...
+#
+
+# NOTE: This script should only run inside a Node.js docker container.
+# See ./load-test.sh for details.
+set -ex
+
+# Create an empty directory to test
+mkdir -p /app && cd /app
+tar zxf /pulsar-client-node/pulsar-client-node.tar.gz
+cp /pulsar-client-node/build/Release/Pulsar.node .
```
shibd, line 28: You can just copy `Pulsar.node` to `/pulsar-client-node/lib/binding`

In this way, does not need to change its: `require()`

tests/load-test.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
shibd, line 1: This script is only used in Linux-related workflows and is recommended to be placed in the `/pkg/linux/` directory

.github/workflows/ci-pr-validation.yml
```diff
@@ -167,6 +167,14 @@ jobs:
           docker run -i -v $PWD:/pulsar-client-node build:latest \
               /pulsar-client-node/pkg/linux/build-napi-inside-docker.sh
 
+      - name: Test NAPI file in other containers
+        if: matrix.image == 'linux_glibc'
+        run: |
+          ./tests/load-test.sh node:16-buster
```
shibd, line 173: Can we keep the param node version consistent with  `${{matrix.nodejs}}`?

pkg/linux/Dockerfile_linux_glibc
```diff
@@ -17,15 +17,19 @@
...
-     apt-get install -y \
-        curl \
-        g++ \
-        make \
-        python3
+ARG ARCH
+WORKDIR /app
+RUN SUFFIX=$(echo ${ARCH} | sed 's/86_//') && \
+  echo $SUFFIX && \
+  curl -O -L https://nodejs.org/download/release/v16.19.0/node-v16.19.0-linux-$SUFFIX.tar.gz && \
```
shibd, line 29: Here need to download the same version of Node.js as defined in [workflow](https://github.com/apache/pulsar-client-node/blob/f5fe0c23a4e6cffe9d781271ea08f181bf5bf6e7/.github/workflows/ci-pr-validation.yml#L135-L136)

shibd (changes requested):

tests/docker-load-test.sh
BewareMyPower: No. If we're doing this way, we need to run `npm install` to install these dependencies: https://github.com/apache/pulsar-client-node/blob/dc2c3c5feb1d51e317215ae183ccf0a1b7306353/package.json#L55-L57

Actually, IMO they are not required. I'm not sure if there is a way to move them from `dependencies` to `devDependencies`. But this script is just for tests and they are equivalent.

tests/load-test.sh
BewareMyPower: `run-unit-tests.sh` is also Linux-related workflows.

https://github.com/apache/pulsar-client-node/blob/dc2c3c5feb1d51e317215ae183ccf0a1b7306353/tests/run-unit-tests.sh#L27

.github/workflows/ci-pr-validation.yml
BewareMyPower: Why? The `load-test.sh` is used to test the `Pulsar.node` works for various versions of Node.js. Did you mean testing Node.js 18 instead of 19?

tests/docker-load-test.sh
shibd: > If we're doing this way, we need to run npm install to install these dependencies

This is ok, this is consistent with how users use.

> And npm install might build the Pulsar.node again on the test docker image, so the Pulsar.node built on CentOS 7 might be overwritten.

When `/pulsar-client-node/lib/binding/Pulsar.node`, It will no longer be compiled and overwritten.

pkg/linux/Dockerfile_linux_glibc
shibd: The GLIBCXX version is too high and is only related to the operating system environment, right? It has nothing to do with the Node.js version, right?

The reason here is consistent with [workflow](https://github.com/apache/pulsar-client-node/blob/f5fe0c23a4e6cffe9d781271ea08f181bf5bf6e7/.github/workflows/ci-pr-validation.yml#L135-L136)  is that the versions we provide to users are all compiled with the same node version (Linux, windows, macOS).

BTW: The current TLS version of Node.js is 18.

.github/workflows/ci-pr-validation.yml
shibd: > The load-test.sh is used to test the Pulsar.node works for various versions of Node.js.

I think that we should test on various types of operating systems(such as centos7, etc). But both use Node.js 18.

tests/docker-load-test.sh
BewareMyPower: > When /pulsar-client-node/lib/binding/Pulsar.node, 

Did you mean when `/pulsar-client-node/lib/binding/Pulsar.node` already exists?

pkg/linux/Dockerfile_linux_glibc
BewareMyPower: >  It has nothing to do with the Node.js version, right?

The pre-built Node.js binary is built with a specific C++ compiler. So there is a GLIBCXX requirement for the `node` binary itself.

You can run the `centos:7` container first:

```bash
docker run --rm -it centos:7 /bin/bash
```

Then run:

```bash
yum update -y
curl -O -L https://nodejs.org/dist/v18.13.0/node-v18.13.0-linux-x64.tar.xz
tar xf node-v18.13.0-linux-x64.tar.xz
./node-v18.13.0-linux-x64/bin/node
```

You will see the following error:

```bash
./node-v18.13.0-linux-x64/bin/node: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by ./node-v18.13.0-linux-x64/bin/node)
./node-v18.13.0-linux-x64/bin/node: /lib64/libc.so.6: version `GLIBC_2.25' not found (required by ./node-v18.13.0-linux-x64/bin/node)
./node-v18.13.0-linux-x64/bin/node: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by ./node-v18.13.0-linux-x64/bin/node)
./node-v18.13.0-linux-x64/bin/node: /lib64/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by ./node-v18.13.0-linux-x64/bin/node)
./node-v18.13.0-linux-x64/bin/node: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by ./node-v18.13.0-linux-x64/bin/node)
./node-v18.13.0-linux-x64/bin/node: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by ./node-v18.13.0-linux-x64/bin/node)
```

Though maybe we can use other ways to install Node.js 18 on CentOS 7, but it could be much more complicated.

We only need to build the `Pulsar.node` with the NAPI of Node.js and the compatibility of NAPI is good. So using Node.js 16 on CentOS 7 is not a problem.

.github/workflows/ci-pr-validation.yml
BewareMyPower: The `node:<version>-<suffix>` image just represents a specific node version and OS. For example, `node:16-buster` represents Node.js 16 and Debian-buster.

CentOS 7 has already been verified when building the `Pulsar.node`. There are many Linux distributions and they cannot be covered all. And it will require some extra efforts to install Node.js on them.

BewareMyPower: Just as the GLIBCXX compatibility issue here https://github.com/apache/pulsar-client-node/pull/288#discussion_r1073622557, installing Node.js 18 on some systems could be very complicated. The only way is to compiling it from source. Since NAPI guarantees the ABI compatibility, why do we insist on verifying with Node.js 18?

tests/load-test.sh
BewareMyPower: IMO, It's good to:
- Put `load_test.js` under `tests/linux`
- Rename `load_test.js` with `linux_load_test.js`

It's bad to put `load_test.js` under `pkg/linux`.

.github/workflows/ci-pr-validation.yml
BewareMyPower: These tests are added to verify `Pulsar.node` can be loaded by Node.js with higher version libstdc++.so dependency.

tests/load-test.sh
shibd: Let's try to summarize these scripts:

- `./tests/run-unit-test.sh`: Used as unit test verification for running CI, suitable for Linux platforms. So, put it in the `/tests` directory.
- `./pkg/load_test.sh`: Used as a validation binary when running CI, suitable for all platforms(Require compile first). So, put it in the `./pkg/` dir.
- `./tests/load-test.sh` and `./tests/docker-load-test.sh`:Used as validation binary on multiple platforms in Linux(Require compile first).

I think `./tests/load-test.sh` and `./tests/docker-load-test.sh` is used to validate multiple platforms and are not part of unit testing.

shibd: @BewareMyPower Can you take a look at it? We want to release this fix to 1.8.1

tests/load-test.sh
```diff
@@ -0,0 +1,43 @@
...
+fi
+IMAGE=$1
+PLATFORM=$2
+
+ROOT_DIR=${ROOT_DIR:-$(git rev-parse --show-toplevel)}
+cd $ROOT_DIR
+
+if [[ ! -f $ROOT_DIR/pulsar-client-node.tar.gz ]]; then
+    git archive -o pulsar-client-node.tar.gz HEAD
+fi
```
RobertIndie, line 40: If this script is not only used in the CI, we need to remove `pulsar-client-node.tar.gz` after executing. If this script is only for CI. I am +1 for moving it  to `pkg` folder.

tests/load-test.sh
BewareMyPower: Actually `load_test.js` should not be named as "xxx_test.js`. It does not "test" anything. It's just a very trivial example script that creates a `Client` object and closes it.

A script that tries to run that script in a specific environment should be treated as "test" (test if `xxx.js` can run with the Node.js client installed in a specific env)

tests/load-test.sh
BewareMyPower: We can run the script locally. For example, assuming there is a new Node.js image (e.g. `node-20-xxx`), you can run `./tests/load-test.sh node-20-xxx x86_64` to verify the `Pulsar.node` built by `pkg/linux/build-napi-inside-docker.sh` can work for that.

Here I don't remove `pulsar-client-node.tar.gz` because it can be reused for later tests.

BewareMyPower: Oh I realized the problem. We should not reuse the `pulsar-client-node.tar.gz` because it might be built with an earlier version. I will modify it.

tests/load-test.sh
shibd: > Actually load_test.js should not be named as "xxx_test.js". It does not "test" anything. It's just a very trivial example script that creates a Client object and closes it.

Yes, It is to test whether the binary package is usable, not unit testing. Similarly, `./tests/load-test.sh` and `docker-load-test.sh`  also test whether the binary package can be used(Multi-platform under Linux). 

They are all part of the compilation process, so I recommend putting them in the `pkg` directory

IMO, That only unit test-related files should be placed in the `tests` directory.

RobertIndie (approved):

tests/load-test.sh
BewareMyPower: > IMO, That only unit test-related files should be placed in the tests directory.

So you mean:
- A `tests` directory should only contain **unit tests**, not other tests.
- A `pkg` directory could contain **tests**, not only packaging scripts.

What makes `pkg` directory so different that it can contain everything?

BewareMyPower: My perspective is, `tests` directory groups all tests into one directory, including:
1. Tests for the `src/` directory.
2. Tests for the `pkg/` directory.
3. ...

If we have some shell scripts in other directories like `bin/`, the tests for them could also be put under the `tests` directory.

Actually I want to move the `load_test.js` into the `tests` directory, but I want to avoid unnecessary project structure changes.

BewareMyPower: It's very clear that:
- `pkg/` directiory contains scripts that package the Node.js binaries in various operation systems.
- `src/` directory containes source code
- `docs/` directory contains documents
- `example/` directory contains examples
- `tests/` directory contains tests

I do not know why do you so insist on distributing tests to different directories.

Some languages tend to put tests under the same directory with the code to test, like Golang, the source code directory contains `xxx.go` and the associated test `xxx_test.go`. It seems that your logic looks like that: the tests (`load-test.sh`) for package scripts should be put under the same directory (`pkg/linux/*.sh`). However, Node.js is different with those languages. Just like Java, the tests for source code (`src/`) are put under a separated directory (`tests/`). If you changed the rule, the for users new to this project, it's not clear to know tests are distributed into how many directories..

BewareMyPower: I don't want to play word games like "verify" and "tests", but since you mentioned this idea, I found a reference here: https://www.quora.com/What-is-the-difference-between-testing-and-verification

> Verification: Before a system is developed, there must be a design where the basic requirements for the system are well spelt out. verification is concerned with whether the system is well-engineered, error-free, and so on. Verification will help to determine whether the software is of high quality.
> Testing refers to the process of executing an application or program with the intent of detecting potential software bugs. It also has the objective of helping the developers to find out whether the software is working according to the intended purpose.

The `load-tests.sh` should be executed **after** the binaries are built and it tests whether the binaries can be executed in a given docker container.

BewareMyPower: I realized https://github.com/apache/pulsar-client-python/blob/main/pkg/test-wheel.sh might inspire you to put scripts under the `pkg/` directory. I don't think it's a good practice.

First, in this script, `/pulsar-client-python/wheelhouse/pulsar_client-*.whl` should be the paths inside a docker container. This script cannot be executed locally. It's more like the `docker-load-test.sh` in this PR.

However, the path of `docker-load-test.sh` is not related to the Dockerfile under `pkg/linux`. Instead, the path of `/pulsar-client-node/build/Release/pulsar.node` and `/pulsar-client-node/pulsar-client-node.tar.gz` are created by `load-test.sh`, not the `pkg/linux/Dockerfile_linux_glibc`.

shibd: Well, In short. My point is to verify `pulsar.node` does not belong to the unit test. Although it is important, it is part of the compilation.

Let's go on and stop tangling.

shibd (approved):

Issue #281, opened 2023-01-06
Failed to create producer: ConnectError when upgrading to 1.8.0


ericallam:
Hi there, I'm just attempting to upgrade to 1.8.0 (from 1.7.0) and I'm now getting a `Failed to create producer: ConnectError` error when creating a producer. If I downgrade back to 1.7.0 the same code works fine. 

```ts
const client = new Pulsar.Client({
  serviceUrl: "pulsar+ssl://<cluster>.<namespace>.snio.cloud:6651",
  authentication: new AuthenticationOauth2({ ... }),
  operationTimeoutSeconds: 5,
});

const producer = await client.createProducer({
  topic: "persistent://public/default/testing-topic",
});
```

I'm running on macOS 12.5.1 arm, Node.js v18.12.1

shibd:
Hi, @ericallam Need to confirm, are you getting an auth `AuthenticationError`? In my environment I reproduce and I get an `AuthenticationError`

I get the error log:

```
Response failed for getting the well-known configuration https://dev-kt-aa9ne.us.auth0.com. Error Code 60: SSL certificate problem: unable to get local issuer certificate
```

You can add the following code to get the log information:
```
  Pulsar.Client.setLogHandler((level, file, line, message) => {
    console.log('[%s][%s:%d] %s', Pulsar.LogLevel.toString(level), file, line, message);
  });
```

shibd:
When I use `pulsar+ssl:`, I get the same result as you: `Error: Failed to create producer: ConnectError`

The essential reason should be this:
```
Response failed for getting the well-known configuration https://dev-kt-aa9ne.us.auth0.com. Error Code 60: SSL certificate problem: unable to get local issuer certificate
```
 I will check it.

ericallam:
Are there any workarounds for this issue @shibd?

ericallam:
Thanks, will take a look at building from source 👍

shibd:
Hi, Let's continue the discussion here.

https://github.com/apache/pulsar-client-node/pull/282#issuecomment-1378497083

I build from source code following Master's `README` and it worked.

I suspect maybe you're using a different server environment?

Can you test by following the steps below:

1. Refer to [How to build](https://github.com/apache/pulsar-client-node/tree/v1.8.0#how-to-build) of master README to build from the source
[...]

PR #300, opened 2023-02-22
[fix] MacOS  tls get certs failed.

shibd:
### Motivation

#281 

MacOS arm64 NAPI is cross-compiled through MacOS x86's runner. The path to `certs` is not automatically recognized during compilation.

Although it is possible to set `tlsTrustCertsFilePath ` in the code to work around, this is not user-friendly.

### Modifications
- Specify the certs path when compiling `curl` and `openSSL`.

### Verifying this change
- I reproduced the issue on my local machine and used the [package](https://github.com/shibd/pulsar-client-node/suites/11135154708/artifacts/567885331) to test the pass.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/pkg/mac/build-cpp-deps-lib.sh
+++ b/pkg/mac/build-cpp-deps-lib.sh
@@ -64,7 +64,7 @@ if [ ! -f openssl-OpenSSL_${OPENSSL_VERSION_UNDERSCORE}.done ]; then
         else
           PLATFORM=darwin64-x86_64-cc
         fi
-        ./Configure --prefix=$PREFIX no-shared no-unit-test $PLATFORM
+        ./Configure --prefix=$PREFIX $SSL_DIR --openssldir=/etc/ssl no-shared no-unit-test $PLATFORM
         make -j8
         make install_sw
     popd
@@ -169,6 +169,7 @@ if [ ! -f curl-${CURL_VERSION}.done ]; then
       CFLAGS="-fPIC -arch ${ARCH} -mmacosx-version-min=${MACOSX_DEPLOYMENT_TARGET}" \
             ./configure --with-ssl=$PREFIX \
               --without-nghttp2 \
+              --with-ca-bundle="/etc/ssl/cert.pem" \
               --without-libidn2 \
               --disable-ldap \
               --without-brotli \
--- a/pkg/mac/build-cpp-deps-lib.sh
+++ b/pkg/mac/build-cpp-deps-lib.sh
@@ -64,10 +64,16 @@ if [ ! -f openssl-OpenSSL_${OPENSSL_VERSION_UNDERSCORE}.done ]; then
         else
           PLATFORM=darwin64-x86_64-cc
         fi
-        ./Configure --prefix=$PREFIX $SSL_DIR --openssldir=/etc/ssl no-shared no-unit-test $PLATFORM
+        ./Configure --prefix=$PREFIX --openssldir=/etc/ssl no-shared no-unit-test $PLATFORM
         make -j8
         make install_sw
     popd
+
+    OPENSSL_DIR=$(./install/bin/openssl version -d | sed 's/\"//g')
+    if [[ $OPENSSL_DIR =~ "\/etc\/ssl" ]]; then
+        echo "The openssl dir is not set as expected: " $OPENSSL_DIR
+        exit 1
+    fi
 
     rm -rf OpenSSL_${OPENSSL_VERSION_UNDERSCORE}.tar.gz openssl-OpenSSL_${OPENSSL_VERSION_UNDERSCORE}
     touch openssl-OpenSSL_${OPENSSL_VERSION_UNDERSCORE}.done
@@ -179,6 +185,12 @@ if [ ! -f curl-${CURL_VERSION}.done ]; then
               --host=$ARCH-apple-darwin
       make -j16 install
     popd
+
+    CERT_PATH=$(./install/bin/curl -v https://example.com 2>&1 | grep CAfile)
+    if [[ $CERT_PATH =~ "\/etc\/ssl" ]]; then
+        echo "The certification path is not set as expected: " $CERT_PATH
+        exit 1
+    fi
 
     rm -rf curl-${CURL_VERSION} curl-${CURL_VERSION}.tar.gz
     touch curl-${CURL_VERSION}.done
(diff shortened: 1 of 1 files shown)


eolivelli:
one thing I don't understand here is that there error was reported by @nicoloboschi on a Intel based Mac and not on an arm64

nicoloboschi (commented):
The fix looks related to darwin in general, so we might need to update the title.

Could we add a simple check in the CI to avoid regressions?

BewareMyPower:
I agree that we might need to add a simple check in CI. I'm not sure if this fix works. Because the SSL handshake error was reported from Boost, not libcurl. I think we need to make Boost depend on a correct OpenSSL to get default verify paths via the `set_default_verify_paths` method:

https://github.com/apache/pulsar-client-cpp/blob/92c182acf7d1212218cb67e315656a23ee32a904/lib/ClientConnection.cc#L216

BewareMyPower:
Boost calls `SSL_CTX_set_default_verify_paths` (from libopenssl) to determine the default CA cert path. I guess it's because the OpenSSL that is built from source does not have the correct path?

shibd:
> one thing I don't understand here is that there error was reported by @nicoloboschi on a Intel based Mac and not on an arm64

@nicoloboschi Can you help confirm this?

And can your pre verify this pr?

1. Download and unzip NAPI.  [MacOS x86_64](https://github.com/shibd/pulsar-client-node/suites/11136563235/artifacts/567980911) or [MacOS arm64](https://github.com/shibd/pulsar-client-node/suites/11136563235/artifacts/567980910)
2. Move `pulsar.node` to your project `node_modules/pulsar-client/lib/binding/`
3. Run you test code.
4. The first time running, you will be receive alerted macOS system. 
image

image

shibd:
> I'm not sure if this fix works. Because the SSL handshake error was reported from Boost, not libcurl. I think we need to make Boost depend on a correct OpenSSL to get default verify paths via the set_default_verify_paths method:

https://github.com/apache/pulsar-client-node/pull/300/files#diff-33eaabd27404f12df463ee1983d2646944310824dcedc87f20e832463fdaef4eR67

@BewareMyPower This change will tell OpenSSL the correct certs path.

nicoloboschi:
> > one thing I don't understand here is that there error was reported by @nicoloboschi on a Intel based Mac and not on an arm64
> 
> @nicoloboschi Can you help confirm this?
> 
> And can your pre verify this pr?
> 
> 1. Download and unzip NAPI.  [MacOS x86_64](https://github.com/shibd/pulsar-client-node/suites/11136563235/artifacts/567980911) or [MacOS arm64](https://github.com/shibd/pulsar-client-node/suites/11136563235/artifacts/567980910)
> 2. Move `pulsar.node` to your project `node_modules/pulsar-client/lib/binding/`
> 3. Run you test code.
> 4. The first time running, you will be receive alerted macOS system.
> 
> image
> 
> image

I confirm this fix is effective and it worked in my tests. Good work! 
I still think that a proper check is needed in the CI

shibd:
> In conclusion, it's a trade-off of the convenience to install the client without building the C++ add-on locally. Users could also solve this issue by copying the cert to a certain directory without any code change.

Agreed, I think we try to match the system to the certification path. If it really doesn't work, the user can solve it by following ways:

- Copy the certificate to the correct path
- Setting the environment variable: 
```
export SSL_CERT_DIR="/etc/ssl/"
export CURL_CA_BUNDLE="/etc/ssl/cert.pem"
```
- Setting the `tlsTrustCertsFilePath ` configuration.

This PR solves the problem with paths under MacOS, and I don't think we should block the release of 1.8.1.

BewareMyPower:
> export CURL_CA_BUNDLE="/etc/ssl/cert.pem"

It does not work because this environment variable only works for the `curl` binary. The libcurl library does not read the environment variable for CA cert store.

I didn't verify `SSL_CERT_DIR` yet, but there is no reason that the libopenssl library will read this environment variable. Even yes, it only specifies the directory, the cert file name could be different. For example, it's `ca-certificates.crt` on Ubuntu and `ca-bundle.crt` on CentOS.

BewareMyPower:
> I don't think we should block the release of 1.8.1

Totally I agree. But the comment [here](https://lists.apache.org/thread/pr9921vflxg96yj0vf870rqptgg05wz1) from @nicoloboschi deserves the discussion.

> I wonder why it wasn't needed in 1.7.0 and now it's required in 1.8.1.
> This can be problematic for users because the client upgrade requires code
> changes and redeployment.

The new way of installing the Node.js client brings the requirement of code changes. I think the only way is to copying the cert file to correct paths. Then I think we must document it well that which cert file path should users configure for these systems:
- Windows
- Linux (including several distributions)
- macOS

It's a breaking change of 1.8.0, we'd better fix it since we've already realized it. If we cannot fix it, we should document the workaround.

eolivelli:
Is there a way to release two versions of the package? Once we statically linked libraries and one with dynamic linking?

shibd:
> Is there a way to release two versions of the package? One with statically linked libraries and one with dynamic linking?

If using dynamic linking, need to install the CPP client first. This falls back to 1.7.0.

BewareMyPower:
> Is there a way to release two versions of the package? One with statically linked libraries and one with dynamic linking?

If we linked to OpenSSL dynamically, we have to link other dependencies dynamically too. It would require users to install all the dependencies (e.g. Boost, Protobuf, etc.) into their systems. At the moment, I cannot think of a way to fix this issue.

I'm wondering if we can get the default CA cert bundle location via the Node.js code and then set it as the default `tlsTrustCertsFilePath` config?

BewareMyPower:
I've tested the Node.js client on Windows, it still cannot find the default CA cert file. But installing the Node.js client 1.7.0 or earlier is really hard and I cannot verify it at the moment.

My conclusion is that it should not block this release if there is no good solution. I'd rather like to document the default CA cert location for different systems.

BewareMyPower:
> I'm wondering if we can get the default CA cert bundle location via the Node.js code and then set it as the default tlsTrustCertsFilePath config? 

Eventually I found Python client fixes this issue with the same way: https://github.com/apache/pulsar-client-python/blob/9b3c55f17adfe6b5a3fa7b016c94d687338c54da/pulsar/__init__.py#L458-L460

Node.js client also provides [an API](https://nodejs.org/api/tls.html#tlsrootcertificates) to use its bundled certificates, instead of using system's certificates. After discussing with @shibd, we're going to adopt this way. So I closed this PR first. After all, we can verify the release by following https://lists.apache.org/thread/xj9ypz8lnoxwgjgks7lsz3hxwvnw5kww

PR #301, opened 2023-02-24
Use the certificate provided by the node.js

shibd:
Master Issue: #281 

### Motivation

After 1.8.0, we helped users precompile CPP, resulting in users not being able to obtain system certificates when using TLS-related functions.

Current usage issues list reference: https://github.com/BewareMyPower/pulsar-tls-examples

Node.js provides [bundled Mozilla CA](https://nodejs.org/api/tls.html#tlsrootcertificates), which we can use by default.

### Modifications
- Use Node.js bundled Mozilla CA when `tlsTrustCertsFilePath` is empty.
- Since the CPP client can't set cert content direct, we first need to write the certificate content to a file and pass in the path to the CPP client.

### Verifying this change

You can clone the [repository](https://github.com/shibd/pulsar-client-tls-test) installation and verify:

```
git clone git@github.com:shibd/pulsar-client-tls-test.git
npm install
node tls-oauth2.js
node tls-token.js
```

The environment I currently verified passed is:
- Windows x64
- MacOS arm64
- Ubuntu 20.04.5

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/GenCertFile.js
+++ b/GenCertFile.js
@@ -19,6 +19,4 @@
 
 const Client = require('./src/Client.js');
 
-(() => {
-  Client.genCertFile();
-})();
+Client.genCertFile();
--- a/GenCertFile.js
+++ b/GenCertFile.js
@@ -17,30 +17,8 @@
  * under the License.
  */
 
-const PulsarBinding = require('./src/pulsar-binding');
-const AuthenticationTls = require('./src/AuthenticationTls.js');
-const AuthenticationAthenz = require('./src/AuthenticationAthenz.js');
-const AuthenticationToken = require('./src/AuthenticationToken.js');
-const AuthenticationOauth2 = require('./src/AuthenticationOauth2.js');
 const Client = require('./src/Client.js');
 
-const LogLevel = {
-  DEBUG: 0,
-  INFO: 1,
-  WARN: 2,
-  ERROR: 3,
-  toString: (level) => Object.keys(LogLevel).find((key) => LogLevel[key] === level),
-};
-
-const Pulsar = {
-  Client,
-  Message: PulsarBinding.Message,
-  MessageId: PulsarBinding.MessageId,
-  AuthenticationTls,
-  AuthenticationAthenz,
-  AuthenticationToken,
-  AuthenticationOauth2,
-  LogLevel,
-};
-
-module.exports = Pulsar;
+(() => {
+  Client.genCertFile();
+})();
--- a/index.js
+++ b/index.js
@@ -22,6 +22,7 @@ const AuthenticationTls = require('./src/AuthenticationTls.js');
 const AuthenticationAthenz = require('./src/AuthenticationAthenz.js');
 const AuthenticationToken = require('./src/AuthenticationToken.js');
 const AuthenticationOauth2 = require('./src/AuthenticationOauth2.js');
+const Client = require('./src/Client.js');
 
 const LogLevel = {
   DEBUG: 0,
@@ -32,7 +33,7 @@ const LogLevel = {
 };
 
 const Pulsar = {
-  Client: PulsarBinding.Client,
+  Client,
   Message: PulsarBinding.Message,
   MessageId: PulsarBinding.MessageId,
   AuthenticationTls,
--- a/index.js
+++ b/index.js
@@ -43,4 +43,8 @@ const Pulsar = {
   LogLevel,
 };
 
+(() => {
+  Client.genCertFile();
+})();
+
 module.exports = Pulsar;
--- a/src/Client.js
+++ b/src/Client.js
@@ -25,7 +25,7 @@ const certsFilePath = `${__dirname}/cert.pem`;
 
 class Client {
   constructor(params) {
-    if (typeof params.tlsTrustCertsFilePath === 'undefined') {
+    if (!params.tlsTrustCertsFilePath) {
       // eslint-disable-next-line no-param-reassign
       params.tlsTrustCertsFilePath = certsFilePath;
     }
--- a/src/Client.js
+++ b/src/Client.js
@@ -1,0 +1,65 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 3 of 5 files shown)


src/Client.js
```diff
@@ -0,0 +1,65 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+const tls = require('tls');
+const fs = require('fs');
+const os = require('os');
+const PulsarBinding = require('./pulsar-binding');
+
+const tmpCertsFilePath = `${__dirname}/cert.pem`;
```
BewareMyPower, line 24: You should add `cert.pem` to the `.gitignore` file.

src/Client.js
```diff
@@ -0,0 +1,65 @@
...
+    if (typeof params.tlsTrustCertsFilePath === 'undefined') {
+      fs.rmSync(tmpCertsFilePath, { force: true });
+      const fd = fs.openSync(tmpCertsFilePath, 'a');
+      try {
+        tls.rootCertificates.forEach((cert) => {
+          fs.appendFileSync(fd, cert + os.EOL);
+        });
+      } finally {
+        fs.closeSync(fd);
+      }
```
BewareMyPower, line 37: There is a problem that each time a `Client` instance is created, the file will be removed, created and written. If multiple `Client` instances are created concurrently, race condition could happen.

I'm wondering if there is a way to initialize the `cert.pem` when running `npm install`. It means that the CA will be created only once when users type `npm install`. It could also ensure the CA is latest and not changed.

BewareMyPower (changes requested):

src/Client.js
BewareMyPower: Here is an example: https://www.npmjs.com/package/node_extra_ca_certs_mozilla_bundle?activeTab=explore

It writes the pem files during `npm install`. You can see the effect that:

```bash
$ npm install node_extra_ca_certs_mozilla_bundle

added 8 packages in 18s

63 packages are looking for funding
  run `npm fund` for details
$ ls node_modules/node_extra_ca_certs_mozilla_bundle/ca_bundle
ca_intermediate_bundle.pem  ca_intermediate_root_bundle.pem  ca_root_bundle.pem
```

shibd: Make sense. I changed it, PTAL.

shibd:
> I suggest we should built-in the cacert. In the `install` step, and the network is well, we can update this cert.

Since Node.js already has built-in certificates, we don't need to make a copy in our repository, we just need to use it when installing.

index.js
```diff
@@ -42,4 +43,8 @@ const Pulsar = {
   LogLevel,
 };
 
+(() => {
```
nodece, line 46: It is better that move the generate cert to new file.

shibd: I want to use `index.js` as some preparation after installation, and if there are other requirements, can add it to this method.

> and this file is pure javascript without native module.
 Generating certificates is part of the node client

nodece: But there's no point in generating it too many times.

shibd: oh, I see. I initially thought this method would only be called once. It turns out that whenever the user `require index.js` it will be executed

BewareMyPower (approved):

GenCertFile.js
```diff
@@ -0,0 +1,24 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+const Client = require('./src/Client.js');
+
+(() => {
+  Client.genCertFile();
+})();
```
nodece, line 24: A simple way:
```
const Client = require('./src/Client.js');

Client.genCertFile();
```

src/Client.js
```diff
@@ -0,0 +1,68 @@
...
+const fs = require('fs');
+const tls = require('tls');
+const os = require('os');
+const PulsarBinding = require('./pulsar-binding');
+
+const certsFilePath = `${__dirname}/cert.pem`;
+
+class Client {
+  constructor(params) {
+    if (typeof params.tlsTrustCertsFilePath === 'undefined') {
```
nodece, line 28:
```suggestion
    if (!params.tlsTrustCertsFilePath) {
```

nodece (approved):
LGTM, Great @shibd

eolivelli:
Great work!

I would like to reiterate that it is better that we don't bundle a CAcert file statically, in that case we would take responsibility on the contents of the files and to maintain the file. This is a burden that the Pulsar project couldn't bear.

The approach on this path is perfecr

src/Client.js
```diff
@@ -0,0 +1,68 @@
...
+  close() {
+    this.client.close();
+  }
+
+  static setLogHandler(params) {
+    PulsarBinding.Client.setLogHandler(params);
+  }
+
+  static genCertFile() {
+    fs.rmSync(certsFilePath, { force: true });
```
eolivelli, line 56: What happens if you bootstrap two clients on the same machine?
Will they use the same file?

Also, we need to set permissions appropriately on this file, otherwise other users may override the contents and lead to a security hole

Cc @michaeljmarshall @lhotari @BewareMyPower

eolivelli (commented):
@shibd can you please confirm that with this patch we are using the cacerts file provided by the user environment and we are not bundling it at build time?

src/Client.js
shibd: > What happens if you bootstrap two clients on the same machine?
Will they use the same file?

The file is generated only once at installation time. 
If pulsar-client is the global install, they will use the same file.
If pulsar-client is not installed globally, each project generates files and uses them separately.

> Also, we need to set permissions appropriately on this file, otherwise other users may override the contents and lead to a security hole

How do I set permissions? I don't think it's necessary, just like the default `/etc/ssl/cert.pem` file in linux, users can override it too, right?

shibd:
> we are using the cacerts file provided by the user environment 

We are using the [Mozilla CA](https://nodejs.org/api/tls.html#tlsrootcertificates) file provided by Node.js

> and we are not bundling it at build time?

Yes, when the user installed `pulsar-client`, we just went to get this certificate content by `tls.rootCertificates` and append a cert file.

BTW: Since the CPP client can't set cert content direct, we first need to write the certificate content to a file and pass in the path to the CPP client.

nodece:
@shibd I think that we need to add the CAcerts notice to the README.md, so like:

### CA certificates
We are using the [Mozilla CA](https://nodejs.org/api/tls.html#tlsrootcertificates) file provided by Node.js

src/Client.js
BewareMyPower: The permission of `ca-certificates.crt` on Linux is:

```bash
$ ls -lh /etc/ssl/certs/ca-certificates.crt
-rw-r--r-- 1 root root 191K Feb  7 15:20 /etc/ssl/certs/ca-certificates.crt
```

But I don't think we need to change the permission. Because this file would be installed under `node_modules/` in the project directory. The owner is who ran `npm install`. It would be that user's fault if the permission was overridden by others.

PR #302, opened 2023-03-03
[fix] Fix error when falling back to local build

massakam:
### Motivation

If fetching pre-built binaries from a remote server fails, node-pre-gyp will try to fallback to building locally. However, even if the Pulsar C++ library is installed locally, this build does not succeed. This is because the C++ library must have been downloaded by a script under the `pkg` directory in this repository. Normal users other than Pulsar developers do not do this.
```sh
$ npm i pulsar-client-1.9.0-rc.0.tgz
...
npm ERR! g++: error: ../pkg/linux/pulsar-cpp/lib/libpulsarwithdeps.a: No such file or directory
```

### Modifications

Fixed `binding.gyp` to try to link to shared libraries installed by package managers such as Homebrew, RPM and APT if no C++ libraries are downloaded under the `pkg` directory.

### Verifying this change

- [x] Make sure that the change passes the CI checks.

### Documentation

- [ ] `doc-required`
- [x] `doc-not-needed`
- [ ] `doc`
- [ ] `doc-complete`

### DIFF
--- a/.github/workflows/ci-pr-validation.yml
+++ b/.github/workflows/ci-pr-validation.yml
@@ -231,3 +231,68 @@ jobs:
       - name: Package Node binaries lib
         run: |
           npx node-pre-gyp package --target_arch=${{ env.TARGET }}
+
+  macos-napi-homebrew:
+    name: Build NAPI macos with CPP lib installed by Homebrew
+    runs-on: macos-latest
+    timeout-minutes: 3000
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v3
+      - name: Use Node.js 18
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: 'npm'
+      - name: Install CPP lib
+        run: |
+          brew install libpulsar
+      - name: Build Node binaries lib
+        run: |
+          npm install --ignore-scripts
+          npx node-pre-gyp configure
+          npx node-pre-gyp build
+      - name: Test loading Node binaries lib
+        run: |
+          node pkg/load_test.js
+      - name: Package Node binaries lib
+        run: |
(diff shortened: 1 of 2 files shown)


equanz (approved):

shibd (changes requested):
@massakam Hi, after v1.8.0, We need to ensure that the release version has the corresponding pre-built binaries.

If the user needs to install a version that has not yet been released, Need to refer to [README#How to build](https://github.com/apache/pulsar-client-node#how-to-build) to build from the source.

This is done to ensure that the CPP version that Node.js client depends on is always correct.

If this PR is merged, this behavior will be changed. If you install `1.9.0-rc.1` (not yet released), the local cpp client version is 2.8.0 (old), and the build may fail.

massakam:
@shibd

> If this PR is merged, this behavior will be changed. If you install `1.9.0-rc.1` (not yet released), the local cpp client version is 2.8.0 (old), and the build may fail.

Hmm, I don't quite understand your concerns... Currently, that build always fails if there is no C++ library under the `pkg` directory, right? When this PR is merged, if there is a library under `pkg` it will be used, otherwise the library installed on the local machine will be used. If the installed library version is high enough, it will succeed. In other words, it increases the chances of a successful build.

Are you concerned that the version of the installed C++ library is not the expected one, resulting in successful compilation but unexpected behavior?

shibd:
> Are you concerned that the version of the installed C++ library is not the expected one, resulting in successful compilation but unexpected behavior?

Yes, That's precisely what I'm worried about. 

I thought for a moment. The PR is a fix for the developer experience. For developers,  we should ensure that the locally installed CPP client is matched.

So, I remove my request change. Thanks for your contribute!

shibd (approved):

PR #307, opened 2023-03-15
[Fix] Wrap c++ exception when creating client.

### LINKED ISSUE #303: libc++abi: terminating with uncaught exception of type std::invalid_argument: The scheme part is missing:
In the latest version of Pulsar Client (8.1).

When I initiate a new PulsarClient, if the service url is an empty string or not a valid url I get the following error and my node application/process stops.

**libc++abi: terminating with uncaught exception of type std::invalid_argument: The scheme part is missing:**

shibd:
Fixes #303 

### Motivation

The `std::invalid_argument` exception causes the node process to exit directly, we should wrap it as a node exception that the user can catch.

### Modifications

- Wrap c++ exception when creating client.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/tests/client.test.js
+++ b/tests/client.test.js
@@ -1,0 +1,46 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+const Pulsar = require('../index.js');
+
+(() => {
+  describe('Client', () => {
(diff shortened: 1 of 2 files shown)


RobertIndie (commented):
It's better to add a test.

I'm curious for the case of empty service URL, why this check doesn't work:
https://github.com/apache/pulsar-client-node/blob/f32953cfd4c2b1b79b7b64ad49aaf51e6c9d3f7b/src/Client.cc#L96-L102

shibd:
> I'm curious for the case of empty service URL, why this check doesn't work:

Because these two judgment conditions are in conflict
```
if (!clientConfig.Has(CFG_SERVICE_URL) || !clientConfig.Get(CFG_SERVICE_URL).IsString()) { 

and 

if (clientConfig.Get(CFG_SERVICE_URL).ToString().Utf8Value().empty()) { 
```

I fixed this, and keeping this allows errors to be checked in advance.

Also, I kept wrapping all the exceptions that occurred when creating the client, so that I could wrap exceptions for other reasons.

tests/client.test.js
```diff
@@ -0,0 +1,46 @@
...
+        })).toThrow('Service URL is required and must be specified as a string');
+      });
+
+      test('Set not string url', async () => {
+        await expect(() => new Pulsar.Client({
+          serviceUrl: -1,
+          operationTimeoutSeconds: 30,
+        })).toThrow('Service URL is required and must be specified as a string');
+      });
+    });
```
RobertIndie, line 44: Need to add another test it failed to set the service URL with the invalid scheme. It will throw an exception from the C++ client side so that we can test the exception wrapper added by this PR.

shibd: Good idea, I added it. PTAL.

RobertIndie (approved):

PR #309, opened 2023-04-04
[Fix] Fix message listener doesn't respect receiver queue size

### LINKED ISSUE #308: [Bug] consumer listener does not respect receiver queue size
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

Pulsar Version: 2.11.0
Pulsar C++ Client Version: 3.1.0 (This is version used by Nodejs 1.8.0 but I do see newer C++ version 3.1.1 available)
Pulsar Nodejs Client Version: 1.8.0

We are using Pulsar Nodejs Client that depends on Pulsar C++ client version (details above)
- When using Consumers with Listener approach, setting "ReceiverQueueSize" is not honored by Consumer
- Receiver Queue Size works when using Receive approach with Consumer

This issue already addressed in Pulsar Java Client:
- We tested using Java Client and noticed that both Listener and Receive honor ReceiverMessageQueue
- Reference of this fix in Java Client: https://github.com/apache/pulsar/issues/11008
- Related PR: https://github.com/apache/pulsar/pull/11455

### Minimal reproduce step

1. Use Shared subscription that has two below consumers: 
1. Start 1st  Consumer with listener callback approach
2. Start 2nd Consumer with Receive option to receive messages
3. Set ReceiverQueueSize for both consumers to: 50
4. Send 1000 messages to topic that Shared Subscription above is subscribed to

For this test Implement Message processing function as below and call the same from both Listener callback and Receive:
ReceiveMessages(msg: any, consumer: any) {
   await delay(500); // 500 ms delay
}

Its same method used by Listener Consumer or Receive Consumer.

### What did you expect to see?

Listener Consumer: ~500 message to be picked up
Receive Consumer: ~500 messages to be picked up

### What did you see instead?

Listener Consumer: You will notice that almost all 1000 or at least more than 900 messages go to Listener
Receive Consumer: Consumer will pick up at most nothing to < 50

### Anything else?

In C++ implementation, it looks like listener callbacks are posted as messages are processed from queue using Boost Async IO library. Once caller posts all messages, new messages are retrieved.

If any receive callback does CPU intensive work, there is delay imposed but if we are doing async/await option in NodeJS - example: waiting for DB call to complete, caller assumes we are done processing with all messages in Receiver Queue and request more. This will look like very fast consumer to broker and all messages are pushed, and Listener Consumer will end up taking up a lot before actually getting any finished. 

Due to this 2nd Consumer with Receive approach barely get any messages.

This issue is well document and addressed in Pulsar Java Client already but similar fix did not make it C++ client?
https://github.com/apache/pulsar/issues/11008

Is this to be considered as behavior by design for C++ client OR this needs to be flagged as bug and needs fix?
Any thoughts, feedback and responses are appreciated. 

Thanks again for your help

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

RobertIndie:
Fixes #308 

### Motivation

This bug is from the nodejs client but not the c++ client. The root cause is that the message listener calls the user callback in an asynchronous way. It will not wait for the result from the user callback, and then it will process the next messages immediately. 

### Modifications

* Use Napi::Promise to wait for the user callback to complete in the message listener.

The message listener can detect whether the user function is asynchronous or synchronous. 
* If it is a synchronous function, it will behave the same as before. This is not an issue because `Napi::Function::Call` will wait for the synchronous function to complete.
* If it is an asynchronous function, the user function must return a `Promise` object. We use `promise.then()` to determine when the user function is finished. The message listener will wait in the C++ client thread. We should not wait in the Node thread because it will block the entire Node main thread. This way, we can also utilize the feature of `MessageListenerThreads`. Previously, the configuration `MessageListenerThreads` did not work for the asynchronous callback. This PR also fixes this issue.

Note: We need to do the same fix for the reader message listener. I will push another PR for it after this PR gets merged.

### Verifying this change

This change added tests.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`


src/Consumer.cc
```diff
@@ -82,9 +96,15 @@ void MessageListener(pulsar_consumer_t *rawConsumer, pulsar_message_t *rawMessag
...
-  MessageListenerProxyData *dataPtr = new MessageListenerProxyData(cMessage, consumer);
+  std::promise<void> promise;
+  std::future<void> future = promise.get_future();
+  MessageListenerProxyData *dataPtr =
+      new MessageListenerProxyData(cMessage, consumer, [&]() { promise.set_value(); });
   listenerCallback->callback.BlockingCall(dataPtr, MessageListenerProxy);
   listenerCallback->callback.Release();
+
+  future.wait();
+  delete dataPtr;
```
BewareMyPower, line 107: It's better not to use `delete` directly, use `std::unique_ptr`.

src/Consumer.cc
```diff
@@ -55,21 +56,34 @@ void Consumer::Init(Napi::Env env, Napi::Object exports) {
...
-      : cMessage(cMessage), consumer(consumer) {}
+  MessageListenerProxyData(std::shared_ptr<pulsar_message_t> cMessage, Consumer *consumer,
+                           std::function<void(void)> callback)
+      : cMessage(cMessage), consumer(consumer), callback(callback) {}
 };
 
 void MessageListenerProxy(Napi::Env env, Napi::Function jsCallback, MessageListenerProxyData *data) {
   Napi::Object msg = Message::NewInstance({}, data->cMessage);
   Consumer *consumer = data->consumer;
-  delete data;
```
BewareMyPower, line 66: I think we should still delete the data. Replace this line with `std::unique_ptr<MessageListenerProxyData*> data_guard` is a common used RAII solution.

src/Consumer.cc
```diff
@@ -82,9 +96,15 @@ void MessageListener(pulsar_consumer_t *rawConsumer, pulsar_message_t *rawMessag
     return;
   }
 
-  MessageListenerProxyData *dataPtr = new MessageListenerProxyData(cMessage, consumer);
+  std::promise<void> promise;
+  std::future<void> future = promise.get_future();
+  MessageListenerProxyData *dataPtr =
+      new MessageListenerProxyData(cMessage, consumer, [&]() { promise.set_value(); });
```
BewareMyPower, line 102:
```suggestion
      new MessageListenerProxyData(cMessage, consumer, [&promise]() { promise.set_value(); });
```

Capture the specific value or reference instead of capturing all values or references.

BewareMyPower (changes requested):

src/Consumer.cc
RobertIndie: The `BlockingCall` only accepts the raw pointer. FYI: https://github.com/nodejs/node-addon-api/blob/main/doc/threadsafe_function.md#blockingcall--nonblockingcall

src/Consumer.cc
RobertIndie: In this PR, the memory allocation and deallocation of `MessageListenerProxyData` are all controlled by the method `MessageListener`. This make the code clearer. And same as above, the blocking call only accepts the raw pointer so we couldn't use unqie_ptr.

src/Consumer.cc
```diff
@@ -55,21 +56,34 @@ void Consumer::Init(Napi::Env env, Napi::Object exports) {
...
 void MessageListenerProxy(Napi::Env env, Napi::Function jsCallback, MessageListenerProxyData *data) {
   Napi::Object msg = Message::NewInstance({}, data->cMessage);
   Consumer *consumer = data->consumer;
-  delete data;
 
   // `consumer` might be null in certain cases, segmentation fault might happend without this null check. We
   // need to handle this rare case in future.
   if (consumer) {
-    jsCallback.Call({msg, consumer->Value()});
+    Napi::Value ret = jsCallback.Call({msg, consumer->Value()});
```
shibd, line 73: >  The root cause is that the message listener calls the user callback in an asynchronous way. It will not wait for the result from the user callback

I have a question, How did you know that this function returned `Napi::Promise`?

RobertIndie: Actually, it doesn't always return `Napi::Promise`. It only returns that when the user functions is an async function. In nodejs, the return value of the async function is `Promise`.

shibd: I see. When using flow code, it will return `Napi::promise`, right?

```node
    subscription: 'sub1',
    subscriptionType: 'Shared',
    ackTimeoutMs: 10000,
    listener: async (msg, msgConsumer) => {
      console.log(msg.getData().toString());
      msgConsumer.acknowledge(msg);
    },
  });
```

RobertIndie: Yes. That's right.

shibd (approved):

src/Consumer.cc
BewareMyPower: I mean use a RAII coding style. For example, if we have a function that only accepts a raw pointer:

```c++
void f(int*);
```

If now we have a raw pointer `p`, we can write:

```c++
std::unique_ptr<int> up{p};  // don't do anything with up
f(p);
// After going out of the scope, `delete p` will be called automatically.
```

If now we have a unique pointer `up`, we can write:

```c++
f(up.get());
// After going out of the scope, `delete up.get()` will be called automatically.
```

src/Consumer.cc
BewareMyPower: We don't need to use `data_guard`, just still use the original pointer. The `unique_ptr` is used to delete the pointer once it's going out of the scope to replace deleting the pointer manually, which could be skipped by an early return.

RobertIndie: So from my understanding, we shouldn't `delete data` here. Right? Otherwise, it will result in double deletion.

BewareMyPower: Yes. Your current code is good.

BewareMyPower (approved):

src/Consumer.cc
```diff
@@ -82,9 +96,14 @@ void MessageListener(pulsar_consumer_t *rawConsumer, pulsar_message_t *rawMessag
     return;
   }
 
-  MessageListenerProxyData *dataPtr = new MessageListenerProxyData(cMessage, consumer);
-  listenerCallback->callback.BlockingCall(dataPtr, MessageListenerProxy);
+  std::promise<void> promise;
+  std::future<void> future = promise.get_future();
+  std::unique_ptr<MessageListenerProxyData> dataPtr = std::unique_ptr<MessageListenerProxyData>(
+      new MessageListenerProxyData(cMessage, consumer, [&promise]() { promise.set_value(); }));
```
BewareMyPower, line 102:
```suggestion
  std::unique_ptr<MessageListenerProxyData> dataPtr(
      new MessageListenerProxyData(cMessage, consumer, [&promise]() { promise.set_value(); }));
```

Simplify the code.

RobertIndie: Looks better. Thanks.

Issue #274, opened 2023-01-03
[feature request] Support batch receive and ack


shibd:
## Motivation

You can refer to [PIP-38](https://github.com/apache/pulsar/wiki/PIP-38%3A-Batch-Receiving-Messages).

## Solution

We already support batch receive in the C++ client https://github.com/apache/pulsar-client-cpp/pull/21,  In the Node.js client side, we can wrap its interface for support.

mguay22:
Hi @shibd - was trying to give this a go and notice there are no methods regarding the new batching implementation in https://github.com/apache/pulsar-client-cpp/blob/main/lib/c/c_Consumer.cc

I was thinking the implementation would follow a similar pattern of calling methods found there in https://github.com/apache/pulsar-client-node/blob/master/src/Client.cc - am I completely off the mark?

Also would this implementation still require the use of `ConsumerReceiveWorker`? Thanks

shibd:
>  was trying to give this a go and notice there are no methods regarding the new batching implementation in https://github.com/apache/pulsar-client-cpp/blob/main/lib/c/c_Consumer.cc

Yes, we need to support the batch receive method on the `c` interface first.

samwaterbury:
I saw that #357 was merged — I'm curious when the next release of the Nodejs Pulsar client might be? Or is there an outstanding todo for batch receive?

Thanks for all your contributions. 🙂

mguay22:
Looks like this just landed in v1.10.0. Thank you! We're eager to try it out and see performance benefits from the batching.

PR #357, opened 2023-12-05
feat: Support batch receive for consumer.

shibd:
Master Issue: #274

### Motivation

#274  Support batch receive for the consumer.

### Modifications

- Add batch receive policy config on the consumer.
- Add batch receive interface on the consumer.
- Refactor consumer `config set` and `verify logic`. 
  - Since config set logic with verify logic for the consumer is separate, it will be more complicated to read the code and subsequent extensions. 

### Verifying this change
- Add batch receive unit test on Consumer.
- Add more config verify unit test on Consumer.

### Documentation

- [x] `doc-required` 

- [ ] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/src/Consumer.cc
+++ b/src/Consumer.cc
@@ -38,6 +38,7 @@ void Consumer::Init(Napi::Env env, Napi::Object exports) {
       DefineClass(env, "Consumer",
                   {
                       InstanceMethod("receive", &Consumer::Receive),
+                      InstanceMethod("batchReceive", &Consumer::BatchReceive),
                       InstanceMethod("acknowledge", &Consumer::Acknowledge),
                       InstanceMethod("acknowledgeId", &Consumer::AcknowledgeId),
                       InstanceMethod("negativeAcknowledge", &Consumer::NegativeAcknowledge),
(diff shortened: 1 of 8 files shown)


src/ConsumerConfig.cc
```diff
@@ -265,6 +291,39 @@ ConsumerConfig::ConsumerConfig(const Napi::Object &consumerConfig, pulsar_messag
     }
     pulsar_consumer_configuration_set_dlq_policy(this->cConsumerConfig.get(), &dlq_policy);
   }
+
+  if (consumerConfig.Has(CFG_BATCH_RECEIVE_POLICY) &&
+      consumerConfig.Get(CFG_BATCH_RECEIVE_POLICY).IsObject()) {
+    Napi::Object propObj = consumerConfig.Get(CFG_BATCH_RECEIVE_POLICY).ToObject();
+    int maxNumMessages = -1;
+    if (propObj.Has(CFG_BATCH_RECEIVE_POLICY_MAX_NUM_MESSAGES) &&
```
gaoran10, line 299: I found Java client will use the `receiverQueueSize` if `maxNumMessages` is bigger than the `receiverQueueSize`, I'm not sure whether we can add this check in the node client.

tests/consumer.test.js
```diff
@@ -315,6 +346,47 @@ const Pulsar = require('../index');
...
+        });
+
+        const consumer = await client.subscribe({
+          topic: topicName,
+          subscription: 'sub1',
+          subscriptionType: 'Shared',
+          batchReceivePolicy: {
+            maxNumMessages: 10,
+            maxNumBytes: -1,
+            timeoutMs: 500,
```
gaoran10, line 363: Could we add a case? Such as sending messages less than 10 to verify timeout works well.

src/Consumer.cc
```diff
@@ -192,36 +193,17 @@ struct ConsumerNewInstanceContext {
...
-    std::string msg("Ack timeout should be 0 or greater than or equal to " +
-                    std::to_string(MIN_ACK_TIMEOUT_MILLIS));
-    deferred->Reject(msg);
-    return deferred->Promise();
-  }
-  int32_t nAckRedeliverTimeoutMs = consumerConfig->GetNAckRedeliverTimeoutMs();
-  if (nAckRedeliverTimeoutMs < 0) {
-    std::string msg("NAck timeout should be greater than or equal to zero");
-    deferred->Reject(msg);
-    return deferred->Promise();
```
BewareMyPower, line 223: Why did you remove the checks here?

shibd: Move to  `consumerConfig->InitConfig` method.

src/ConsumerConfig.h
```diff
@@ -21,14 +21,17 @@
...
 #include "MessageListener.h"
 
 #define MIN_ACK_TIMEOUT_MILLIS 10000
 
 class ConsumerConfig {
  public:
-  ConsumerConfig(const Napi::Object &consumerConfig, pulsar_message_listener messageListener);
+  ConsumerConfig();
   ~ConsumerConfig();
+  void InitConfig(const std::shared_ptr<ThreadSafeDeferred> deferred, const Napi::Object &consumerConfig,
```
BewareMyPower, line 33: Don't pass const value as the parameter type. For example,

```c++
struct Foo {
  void f1() const {}
  void f2() {}
};

void f(const std::shared_ptr<Foo> foo) {
  foo->f1();
  foo->f2();  // OK, even though f2() is a non-const method
}
```

`f()` can only guarantee `foo` itself cannot be modified like `foo.set(nullptr)`.

However, since this method only needs a reference to the `ThreadSafeDeferred` object and does not transfer or share the ownership of `deferred` to another thread, I suggest passing a reference here.

```c++
  void InitConfig(ThreadSafeDeferred &deferred, const Napi::Object &consumerConfig,
                  pulsar_message_listener messageListener);
```

Then call `deferred.Reject` instead of `deferred->Reject` in the implementation.

BTW, I still don't get why this refactoring is needed in this PR.

tests/consumer.test.js
BewareMyPower: +1. You'd better test the `timeoutMs` config works.

src/Consumer.cc
shibd: Sorry for the late reply. The original code, check settings logic are separated, which will have the following problems:

1. For Some settings, we need to call the C API to determine whether it is successful or not, such as https://github.com/apache/pulsar-client-node/pull/357/files#diff-ca2d7e89de2051db6eacdda40a5abc459724a390b40df60a1e54857caa2ee5f5R321-R325, which we can't do with the original code.
2. After refactoring this way, putting the check settings in a method will make the code clear.

src/ConsumerConfig.h
shibd: Got it. Thanks for your reminder.

shibd: I will keep use std::shared_ptr<ThreadSafeDeferred>, because it will create by NAPI method: `auto deferred = ThreadSafeDeferred::New(info.Env());`

src/ConsumerConfig.cc
shibd: This logic has been added on the CPP client:  https://github.com/apache/pulsar-client-cpp/blob/main/lib/ConsumerImplBase.cc#L38-L49

tests/consumer.test.js
shibd: I will add it. 

BTW: we shouldn't need to do tests with very complete boundaries on the node client, because it depends on the CPP client, and we've added tests for it in cpp

src/ConsumerConfig.cc
gaoran10: Got it, thanks.

gaoran10 (approved):

BewareMyPower (approved):

PR #406, opened 2025-02-28
Fix message might lost when use listener

shibd:
### Motivation

When using a listener to create a consumer with multiple topics, messages might be lost.

After #219, when a message reaches a listener, if the consumer is null, it will ignore this msg, and then msg will be lost.
https://github.com/apache/pulsar-client-node/blob/36b154eadea1d08eb657b541bbeb18f140889526/src/Consumer.cc#L80-L84

Question: Why is the consumer null here?

In CPP implementation, after https://github.com/apache/pulsar-client-cpp/pull/447, we will make sure all sub-consumers create success and then resumeListenerDispacther.

And it will call `create-callback` first
https://github.com/apache/pulsar-client-cpp/blob/54e529aaf82bddac063c847d4c11d3fba3acf0f3/lib/MultiTopicsConsumerImpl.cc#L151-L156
```
            multiTopicsConsumerCreatedPromise_.setValue(get_shared_this_ptr());
            // Now all child topics are successfully subscribed, start messageListeners
            if (messageListener_ && !conf_.isStartPaused()) {
                LOG_INFO("Start messageListeners");
                resumeMessageListener();
            }
```

And, when callback reach to Node.js lib. will execute code: this code will set `consumer object`.

https://github.com/apache/pulsar-client-node/blob/36b154eadea1d08eb657b541bbeb18f140889526/src/Consumer.cc#L176-L189

However, since the consumer is set in `deferred->Resolve`, which is asynchronous, it might not be fully set up when the listener starts receiving messages. Therefore, in the code above, the consumer might be `null`, and the message could ultimately be ignored.

### Modifications

- Use consumer future to make sure get consumer point after setting it.

### Verifying this change
This issue to hard use unit test to cover it.

There is reproduction step:

1. Create a topic with 5 partition.
2. Create subscription
3. Change pulsar source code liks below. (Slow down the subscription to one of the topics to delay the overall completion of the consumer subscription.)
```diff
    protected void handleSubscribe(final CommandSubscribe subscribe) {
        checkArgument(state == State.Connected);
        final long requestId = subscribe.getRequestId();
        final long consumerId = subscribe.getConsumerId();
        TopicName topicName = validateTopicName(subscribe.getTopic(), requestId, subscribe);
+        if (!subscribe.getTopic().equals("persistent://public/default/test-listener2-partition-0")) {
+            try {
+              log.info("test sleep 5000ms: " + subscribe.getTopic());
+               Thread.sleep(5000);
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+        }
```
4. Send 2 msg to `partition-0`
5. Create a consumer with a listener. 
```
  // Create a consumer with listener
    topic: 'persistent://public/default/test-listener2',
    subscription: 'sub1',
    subscriptionType: 'Shared',
    ackTimeoutMs: 100000000,
    listener: (msg, msgConsumer) => {
      console.log("Receive-msg: " + msg.getData().toString());
    },
  });
```

When running repeatedly, you'll encounter receiving only one message or no messages approximately every 2 to 3 times.

**!!Note: After this PR, running it 100 times repeatedly no longer results in any issues.**

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/src/Consumer.cc
+++ b/src/Consumer.cc
@@ -77,30 +77,28 @@ void MessageListenerProxy(Napi::Env env, Napi::Function jsCallback, MessageListe
   Napi::Object msg = Message::NewInstance({}, data->cMessage);
   Consumer *consumer = data->consumer;
 
-  if (consumer) {
-    Napi::Value ret;
-    try {
-      ret = jsCallback.Call({msg, consumer->Value()});
-    } catch (std::exception &exception) {
-      logMessageListenerError(consumer, exception.what());
-    }
+  Napi::Value ret;
+  try {
+    ret = jsCallback.Call({msg, consumer->Value()});
+  } catch (std::exception &exception) {
+    logMessageListenerError(consumer, exception.what());
+  }
 
-    if (ret.IsPromise()) {
-      Napi::Promise promise = ret.As<Napi::Promise>();
-      Napi::Function catchFunc = promise.Get("catch").As<Napi::Function>();
+  if (ret.IsPromise()) {
+    Napi::Promise promise = ret.As<Napi::Promise>();
+    Napi::Function catchFunc = promise.Get("catch").As<Napi::Function>();
 
-      ret = catchFunc.Call(promise, {Napi::Function::New(env, [consumer](const Napi::CallbackInfo &info) {
-                             Napi::Error error = info[0].As<Napi::Error>();
-                             logMessageListenerError(consumer, error.what());
-                           })});
+    ret = catchFunc.Call(promise, {Napi::Function::New(env, [consumer](const Napi::CallbackInfo &info) {
+                           Napi::Error error = info[0].As<Napi::Error>();
+                           logMessageListenerError(consumer, error.what());
+                         })});
 
-      promise = ret.As<Napi::Promise>();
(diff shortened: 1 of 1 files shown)


src/Consumer.cc
```diff
@@ -168,23 +166,13 @@ struct ConsumerNewInstanceContext {
     auto cConsumer = std::shared_ptr<pulsar_consumer_t>(rawConsumer, pulsar_consumer_free);
     auto listener = consumerConfig->GetListenerCallback();
 
-    if (listener) {
```
shibd, line 171: We do not this after https://github.com/apache/pulsar-client-cpp/pull/447

BewareMyPower (approved):

src/Consumer.cc
BewareMyPower: Could you open an issue for it?

shibd: It has already been removed in this PR.

BewareMyPower: On I see. I misunderstood it.

PR #410, opened 2025-03-26
Support set connectionTimeout param

shibd:
### Motivation

In the C++ client, the `connectionTimeout` parameter controls the time allowed to establish a connection, with a default value of 10 seconds

https://github.com/apache/pulsar-client-cpp/blob/4ba83e83fbb4319c0b4cda82372caf042e9ccaa6/lib/ClientConnection.cc#L184-L185

We need to expose this config on node.js client.

### Modifications
- The pulsar_client structure definition from the C++ client has been copied, allowing node.js client to set it using the C++ API instead of the C API.
- Expose `connectionTimeoutMs` configuration

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -33,6 +33,7 @@ export interface ClientConfig {
   listenerName?: string;
   log?: (level: LogLevel, file: string, line: number, message: string) => void;
   logLevel?: LogLevel;
+  connectionTimeoutMs?: number;
 }
 
 export class Client {
--- a/src/Client.cc
+++ b/src/Client.cc
@@ -26,6 +26,7 @@
 #include <pulsar/c/client.h>
 #include <pulsar/c/client_configuration.h>
 #include <pulsar/c/result.h>
+#include "pulsar/ClientConfiguration.h"
 
 static const std::string CFG_SERVICE_URL = "serviceUrl";
 static const std::string CFG_AUTH = "authentication";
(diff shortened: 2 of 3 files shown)


hrsakai:
This fix itself seems fine, but are there any concerns about using the C++ API in other places?
I'm not sure if the C++ API should be allowed.

BewareMyPower:
@hrsakai C and C++ API are released as a whole for source code and pre-built binaries. I think the Node.js client uses the C API due to historical reasons. Using C++ API should be okay. They make no difference from a perspective of Node.js client users.

hrsakai:
@BewareMyPower 
Isn't there a possibility that bugs related to object life cycles may occur if we continue to use C++ APIs in the future?

BewareMyPower:
> that bugs related to object life cycles may occur

We already have a handle leak check here: https://github.com/apache/pulsar-client-node/blob/b79ead00441779a348e921f1aa638f1e2dce35c9/package.json#L23, and https://github.com/apache/pulsar-client-node/pull/412 is an example to show how the memory leak is detected.

The C API just wraps the C++ API, so it should not be expected if we turn to the C++ API directly.

hrsakai (approved):
LGTM

(Since the C++ API and the C API will be mixed, I would like to consider refactoring it in the future.)

PR #416, opened 2025-05-21
[fix][client] handle subscriptionRolePrefix properly

### LINKED ISSUE #415: subscriptionRolePrefix is not handled properly in ReaderConfig
In the following section of `src/ReaderConfig.cc`:

```cpp
if (readerConfig.Has(CFG_SUBSCRIPTION_ROLE_PREFIX) &&
      readerConfig.Get(CFG_SUBSCRIPTION_ROLE_PREFIX).IsString()) {
    std::string subscriptionRolePrefix =
        readerConfig.Get(CFG_SUBSCRIPTION_ROLE_PREFIX).ToString().Utf8Value();
    if (!subscriptionRolePrefix.empty())
      pulsar_reader_configuration_set_reader_name(this->cReaderConfig.get(), subscriptionRolePrefix.c_str());
  }
```

it should call `pulsar_reader_configuration_set_subscription_role_prefix` instead of `pulsar_reader_configuration_set_reader_name`.

This bug results in warnings such as:

```
[17:03:34.537] WARN (app logger/21232 on M-01390): Pulsar warning: [192.168.0.124:50068 -> 54.75.236.18:6651] Received error response from server: AuthorizationError (org.apache.pulsar.broker.PulsarServerException: Failed to create consumer - The subscription name needs to be prefixed by the authentication role, like 99b03f54-91fd-41c0-8e8f-0c43006cbe8c-xxxx for topic: persistent://prod/LEGO.ProductDevelopment.ProductDiscovery/products-b778c389-eba6-43f3-b665-433839b40194-v1) -- req_id: 0
```

which, depending on the Pulsar cluster setup, can result in unexpected behaviour or authorization errors.

candu:
Fixes #415 

### Motivation

At my current employer, we have a centrally-managed Pulsar cluster that requires use of a pre-assigned `subscriptionRolePrefix` for your client when creating a reader.

As per docs, you should be able to do this with:

```
client.createReader({
  topic,
  startMessageId: Pulsar.MessageId.earliest(),
  subscriptionRolePrefix: subscriptionPrefix,
});
```

However, this bug prevents that from working correctly.

### Modifications

Small change in `ReaderConfig.cc` to call the correct `PULSAR_PUBLIC` function.

### Verifying this change

This change is a trivial rework / code cleanup without any test coverage.  That said: I did test that, by using my patched client, I can create a reader in our internal setup described above.  (Without the patch, it fails.)

(Due to NDA, I can't describe that setup in more detail, sadly.)

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

Docs are not needed: this fixes a feature that is already documented, but which did not previously work as intended.

### DIFF
--- a/src/ReaderConfig.cc
+++ b/src/ReaderConfig.cc
@@ -72,7 +72,8 @@ ReaderConfig::ReaderConfig(const Napi::Object &readerConfig, pulsar_reader_liste
     std::string subscriptionRolePrefix =
         readerConfig.Get(CFG_SUBSCRIPTION_ROLE_PREFIX).ToString().Utf8Value();
     if (!subscriptionRolePrefix.empty())
-      pulsar_reader_configuration_set_reader_name(this->cReaderConfig.get(), subscriptionRolePrefix.c_str());
+      pulsar_reader_configuration_set_subscription_role_prefix(this->cReaderConfig.get(),
+                                                               subscriptionRolePrefix.c_str());
   }
 
   if (readerConfig.Has(CFG_READ_COMPACTED) && readerConfig.Get(CFG_READ_COMPACTED).IsBoolean()) {


shibd (approved):
Nice catch!

PR #433, opened 2025-09-25
Support custome router for producer

shibd:
### Motivation

To support a custom router in the Node.js client, the most direct approach would be to call the user-provided router function from within the C++ callback. However, because the C++ client's `sendAsync` method has a synchronous component for partition selection before it becomes fully asynchronous, this implementation causes a deadlock between the Node.js main thread and the C++ callback thread. This is due to the single-threaded nature of the Node.js event loop.

https://github.com/apache/pulsar-client-cpp/blob/90ea3695b80660f837785d98e96047d90de3f64f/lib/PartitionedProducerImpl.cc#L209-L220

You can understand by reviewing the [PR](https://github.com/shibd/pulsar-client-node/pull/44) and running the associated tests.

### Modifications

This PR achieves the goal by calling the user's router function at the **JS layer** and then passing the resulting partition index to the C++ layer via message properties.

1.  The `send` method in `Producer.js` is wrapped. It now first calls the user's router function to asynchronously determine the target partition.
2.  The resulting partition index is added to the message's `properties` under the special key: `__partition__`.
3.  Inside the N-API implementation, a fixed, synchronous C++ function (`internalCppMessageRouter`) is now used, which simply reads this property and returns the partition index to the underlying C++ client.

#### Points for Discussion

This implementation introduces two points for discussion:

1.  We need to maintain a cache for the number of partitions (`numPartitions`) in `Producer.js`, which is then passed to the user's router function. (Currently, this cache has a 60-second TTL. A potential future improvement is to expose a method from the C++ producer to get the partition count directly. This would allow us to remove the JS-side cache, as the C++ client internally handles partition metadata updates.)
2.  This introduces a special property (`__partition__`) in the message properties, which could potentially be confusing for users. However, I believe this is not a major issue at the moment.

### Verifying this change
- The `Message Routing` test case has been added to cover this change.

### Documentation

- [ ] `doc-required`

- [x] `doc-not-needed`
(Please explain why)

- [ ] `doc`

- [ ] `doc-complete`


tests/producer.test.js
```diff
@@ -156,5 +159,95 @@ const Pulsar = require('../index');
...
+            method: 'PUT',
+          },
+        );
+        // 204 No Content is success for PUT create
+        expect(createPartitionedTopicRes.statusCode).toBe(204);
+
+        // 2. Create a producer with the custom message router
+        const producer = await client.createProducer({
+          topic: partitionedTopic, // Note: For producer, use the base topic name
+          messageRouter: (message, topicMetadata) => targetPartition,
```
BewareMyPower, line 188: It's better to show the router can get the message's key or value, as well as the number of partitions from `topicMetadata`, otherwise the parameters will be meaningless

shibd: Good point, I implemented a simple hash method and use message and topicMetadata.

shibd:
Support on PR: https://github.com/apache/pulsar-client-node/pull/435

PR #445, opened 2025-12-17
Support end to end encryption/decryption

shibd:
### Motivation
Support end to end encryption/decryption

### Modifications
- Support end to end encryption/decryption
- Support getEncryptionContext by message

### Verifying this change
- Add a end to end encryption unit test to cover this change.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -333,6 +333,7 @@ export type HashingScheme =
   'JavaStringHash';
 
 export type CompressionType =
+  'None' |
   'Zlib' |
   'LZ4' |
   'ZSTD' |
--- a/src/Message.cc
+++ b/src/Message.cc
@@ -23,6 +23,7 @@
 #include <pulsar/Message.h>
 #include <pulsar/MessageBuilder.h>
 #include <pulsar/EncryptionContext.h>
+#include <map>
 
 struct _pulsar_message {
   pulsar::MessageBuilder builder;
@@ -39,6 +40,12 @@ static const std::string CFG_DELIVER_AFTER = "deliverAfter";
 static const std::string CFG_DELIVER_AT = "deliverAt";
 static const std::string CFG_DISABLE_REPLICATION = "disableReplication";
 static const std::string CFG_ORDERING_KEY = "orderingKey";
+
+static const std::map<pulsar::CompressionType, std::string> COMPRESSION_TYPE_MAP = {
+    {pulsar::CompressionNone, "None"},     {pulsar::CompressionLZ4, "LZ4"},
+    {pulsar::CompressionZLib, "Zlib"},     {pulsar::CompressionZSTD, "ZSTD"},
+    {pulsar::CompressionSNAPPY, "SNAPPY"},
+};
 
 Napi::FunctionReference Message::constructor;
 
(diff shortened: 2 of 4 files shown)


src/CryptoKeyReader.h
```diff
@@ -0,0 +1,39 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef CRYPTO_KEY_READER_H
+#define CRYPTO_KEY_READER_H
+
+#include <napi.h>
+#include <pulsar/CryptoKeyReader.h>
+#include <thread>
```
BewareMyPower:
```suggestion
```

src/CryptoKeyReader.cc
```diff
@@ -0,0 +1,182 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "CryptoKeyReader.h"
+#include <pulsar/Result.h>
+#include <pulsar/EncryptionKeyInfo.h>
+#include <thread>
+#include <future>
+#include <iostream>
```
BewareMyPower:
```suggestion
```

src/CryptoKeyReader.cc
```diff
@@ -0,0 +1,182 @@
...
+    return executeCallback("getPublicKey", keyName, metadata, encKeyInfo);
+  }
+
+  pulsar::Result getPrivateKey(const std::string& keyName, std::map<std::string, std::string>& metadata,
+                               pulsar::EncryptionKeyInfo& encKeyInfo) const override {
+    return executeCallback("getPrivateKey", keyName, metadata, encKeyInfo);
+  }
+
+ private:
+  mutable Napi::ObjectReference jsObject_;
```
BewareMyPower: Instead of adding `mutable` here, it's better to change `CallbackData.jsObjectRef` to `const Napi::ObjectReference*` because only const methods are called on `jsObjectRef`.

index.d.ts
```diff
@@ -198,6 +202,22 @@ export interface TopicMetadata {
  */
 export type MessageRouter = (message: Message, topicMetadata: TopicMetadata) => number;
 
+export interface EncryptionKey {
+  key: string;
+  value: string;
```
BewareMyPower: The value should be a byte array. In C++, `std::string` could also used as a byte array, but in Node.js, `Buffer` should be used as a byte array.

Sorry for not documenting which fields should represent byte array in `EncryptionContext`, but I just checked the implementation and confirmed the following fields should be bytes rather than strings:
- `EncryptionKey::value`
- `param`

index.d.ts
```diff
@@ -198,6 +202,22 @@ export interface TopicMetadata {
...
+  metadata: { [key: string]: string };
+}
+
+export interface EncryptionContext {
+  keys: EncryptionKey[];
+  param: Buffer;
+  algorithm: string;
+  compressionType: CompressionType;
+  uncompressedMessageSize: number;
+  batchSize: number;
```
BewareMyPower: These fields are not tested, you can create a producer with compression and batching enabled to verify they're converted correctly from the C++ result.

PR #453, opened 2026-01-28
Fix issue #3505: Set client description to identify Node.js client

shibd:
## Summary
This PR fixes issue #3505 by adding a `description` parameter to the Node.js client configuration. The default description is set to `"node"`, which causes the client version in topic stats to display as `"Pulsar-CPP-vX.Y.Z-node"` instead of just `"Pulsar-CPP-vX.Y.Z"`.

## Problem
The Node.js client wraps the C++ client library. When connecting to the broker, it uses the C++ client's version string, making it impossible to distinguish Node.js clients from C++ clients in topic statistics. This creates confusion for operators monitoring their Pulsar clusters.

## Solution
- Added `description` parameter to `ClientConfig` interface
- Set default description to `"node"` to identify the client as Node.js
- Allow users to override the description if needed
- Utilize the existing `setDescription()` method from the C++ `ClientConfiguration` class

## Changes
- `src/Client.cc`: Add description configuration parsing and set default value
- `index.d.ts`: Add `description?: string` to `ClientConfig` interface

## Testing
After applying this fix, the `clientVersion` in topic stats will display as:
- With default: `"Pulsar-CPP-v4.0.0-node"`
- With custom description: `"Pulsar-CPP-v4.0.0-node-v1.0.0"` (if description is `"node-v1.0.0"`)

## Backward Compatibility
This change is fully backward compatible. Existing code will continue to work, and the client version string in topic stats will simply include the additional `-node` suffix.

## Related Issues
Fixes #3505

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -27,7 +27,10 @@
       "include_dirs": [
         "<!@(node -p \"require('node-addon-api').include\")",
       ],
-      "defines": ["NAPI_VERSION=4"],
+      "defines": [
+        "NAPI_VERSION=4",
+        "PULSAR_CLIENT_NODE_VERSION=\"<!(node -e \\\"require('./package.json').version\\\")\""
+      ],
       "sources": [
         "src/addon.cc",
         "src/Message.cc",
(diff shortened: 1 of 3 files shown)


shibd:
Replacing with updated PR that correctly implements the fix - automatically setting client description without requiring user configuration.

PR #481, opened 2026-04-22
feat: add replicateSubscriptionState to ConsumerConfig

### LINKED ISSUE #478: Add replicateSubscriptionState to ConsumerConfig
**Summary**

The ConsumerConfig interface does not expose `replicateSubscriptionState`, which is needed for geo-replication failover scenarios where subscription cursor state should be synchronized across clusters.

**Context**

The underlying C++ client supports `replicateSubscriptionStateEnabled` as a consumer configuration option, and it is exposed in other clients.  The Node.js client wraps the C++ library, so the capability should be available at the binding layer, it just needs to be surfaced in the ConsumerConfig interface and passed through to the C++ ConsumerConfiguration.

**Expected API**

```
export interface ConsumerConfig {
    // ... existing properties ...
    replicateSubscriptionState?: boolean;
}
```

When set to true, the broker synchronizes acknowledgment state for this subscription across geo-replicated clusters, enabling seamless consumer failover without reprocessing the backlog.

**Use Case**

When deploying Pulsar with geo-replication across multiple regions for active-passive consumer failover, standby consumers in the secondary region to need to have a warm subscription cursor that reflects what the primary region has already processed. Without `replicateSubscriptionState`, the only options are reprocessing the full backlog on failover or building custom cursor synchronization.

shibd:
## Motivation

Fixes #478

The `ConsumerConfig` interface was missing `replicateSubscriptionState`, which is needed for geo-replication failover scenarios where subscription cursor state should be synchronized across clusters.

## Changes

- Added `CFG_REPLICATE_SUBSCRIPTION_STATE` constant in `src/ConsumerConfig.cc`
- Added handling to pass the config value through to `ConsumerConfiguration::setReplicateSubscriptionStateEnabled()` on the underlying C++ object
- Added `replicateSubscriptionState?: boolean` to the `ConsumerConfig` TypeScript interface in `index.d.ts`
- Added a test case in `tests/consumer.test.js` to verify the option can be set when creating a consumer

### DIFF
--- a/build-support/download-release-artifacts.py
+++ b/build-support/download-release-artifacts.py
@@ -53,6 +53,13 @@ for artifact in data['artifacts']:
     name = artifact['name']
     url = artifact['archive_download_url']
 
+    # docker/build-push-action uploads build records as auxiliary artifacts.
+    # They are not release packages and break the staging flow if we try to
+    # unpack them alongside the platform tarballs.
+    if name.endswith('.dockerbuild'):
+        print(f'Skipping auxiliary artifact {name}')
+        continue
+
     print(f'Downloading {name} from {url}')
     artifact_response = requests.get(url, headers=headers, stream=True)
     artifact_response.raise_for_status()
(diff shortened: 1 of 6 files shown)


shibd:
Closing in favor of a cleaner PR based on the latest master branch.

PR #493, opened 2026-06-23
[feat] Expose message replication metadata

### LINKED ISSUE #480: Expose isReplicated() / getReplicatedFrom() on the Message object
## Feature Request

Expose replication-origin metadata on the Node.js `Message` object so consumer logic can detect whether a received message was produced locally or geo-replicated from a remote cluster.

### Requested API additions

```typescript
// TypeScript definition
interface Message {
  // ... existing methods ...

  /**
   * Returns true if this message was geo-replicated from another cluster.
   */
  isReplicated(): boolean;

  /**
   * Returns the name of the cluster this message was replicated from,
   * or an empty string if the message was produced locally.
   */
  getReplicatedFrom(): string;
}
```

### Implementation

The `replicated_from` field is part of the Pulsar binary protocol's `MessageMetadata` and is already tracked internally. `Message.cc` already accesses the underlying C++ message object directly (see `GetEncryptionContext`), so the same pattern applies:

```cpp
// Message.h — add to private methods:
Napi::Value GetIsReplicated(const Napi::CallbackInfo &info);
Napi::Value GetReplicatedFrom(const Napi::CallbackInfo &info);

// Message.cc — Init, add to DefineClass:
InstanceMethod("isReplicated", &Message::GetIsReplicated),
InstanceMethod("getReplicatedFrom", &Message::GetReplicatedFrom),

// Message.cc — implementations:
Napi::Value Message::GetIsReplicated(const Napi::CallbackInfo &info) {
  Napi::Env env = info.Env();
  if (!ValidateCMessage(env)) return env.Null();
  return Napi::Boolean::New(env, this->cMessage.get()->message.isReplicated());
}

Napi::Value Message::GetReplicatedFrom(const Napi::CallbackInfo &info) {
  Napi::Env env = info.Env();
  if (!ValidateCMessage(env)) return env.Null();
  return Napi::String::New(env, this->cMessage.get()->message.getReplicatedFrom());
}
```

### Dependency

This depends on `isReplicated()` and `getReplicatedFrom()` being added to the C++ client's public `Message` API: apache/pulsar-client-cpp#569.

### Related

- apache/pulsar-client-node#478 — `replicateSubscriptionState` support (also needed for geo-replication failover topologies)

shibd:
### What changed

- Expose `Message.isReplicated()` and `Message.getReplicatedFrom()` from the Node.js binding.
- Add TypeScript definitions for both methods.
- Assert the default non-replicated message behavior in the end-to-end test and type-check sample.

### Why

Consumers need access to Pulsar replication-origin metadata to distinguish locally produced messages from messages geo-replicated from another cluster. This implements apache/pulsar-client-node#480 on top of the C++ client API added in apache/pulsar-client-cpp#583.

`isReplicated()` is derived from whether `getReplicatedFrom()` returns a non-empty cluster name.

Fixes #480

### Validation

- `npm run type-check`
- `npx clang-format-lint src/Message.cc src/Message.h`
- `npx node-pre-gyp configure build`
- `node -e "const Pulsar=require('./'); console.log(typeof Pulsar.Message.prototype.isReplicated, typeof Pulsar.Message.prototype.getReplicatedFrom)"`
- `npx eslint tests/end_to_end.test.js` (passes with existing warnings)

### DIFF
--- a/src/Message.cc
+++ b/src/Message.cc
@@ -63,6 +63,8 @@ Napi::Object Message::Init(Napi::Env env, Napi::Object exports) {
        InstanceMethod("getPartitionKey", &Message::GetPartitionKey),
        InstanceMethod("getOrderingKey", &Message::GetOrderingKey),
        InstanceMethod("getProducerName", &Message::GetProducerName),
+       InstanceMethod("isReplicated", &Message::GetIsReplicated),
+       InstanceMethod("getReplicatedFrom", &Message::GetReplicatedFrom),
        InstanceMethod("getEncryptionContext", &Message::GetEncryptionContext)});
 
   constructor = Napi::Persistent(func);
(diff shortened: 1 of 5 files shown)


src/Message.cc
```diff
@@ -172,6 +174,29 @@ Napi::Value Message::GetProducerName(const Napi::CallbackInfo &info) {
   return Napi::String::New(env, pulsar_message_get_producer_name(this->cMessage.get()));
 }
 
+Napi::Value Message::GetIsReplicated(const Napi::CallbackInfo &info) {
```
BewareMyPower: It would be better to name it as `IsReplicated`.

BewareMyPower (approved):

